PR #11, opened 2020-06-03
[WIP] beginning of dev playbook with pre_tasks to mount volume

cat-bro:
This is a work in progress.

### DIFF
--- a/dev_playbook.yml
+++ b/dev_playbook.yml
@@ -1,0 +1,25 @@
+- hosts: galaxyservers
+  become: true
+  pre_tasks:
+    - name: Detect attached volume
+      parted:
+        device: /dev/vdb
+        unit: GiB
+      register: volume_info
+    - debug:
+        msg: Detected {{ volume_info }}
+    - debug:
+        msg: Everything's gonna be alright
+      when: volume_info.disk is defined and volume_info.disk.dev is defined and volume_info.disk.dev == '/dev/vdb'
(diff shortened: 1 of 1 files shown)


cat-bro:
The volume_info returned from the `parted` step is 
```
{u'disk': {u'dev': u'/dev/vdb', u'physical_block': 512, u'table': u'unknown', u'logical_block': 512, u'model': u'Virtio Block Device', u'unit': u'gib', u'size': 200.0}, 'failed': False, u'partitions': [], u'changed': False, u'script': u\"unit 'GiB' print\"}
```

cat-bro:
@Nick-Rhodes-QFAB I'm not that familiar with the operations here.  Do the parameters look reasonable?  The part of the script I have commented out is the part I have not run yet.

By default the filesystem module will not create a filesystem on a device that already has one.  The when clause should stop it from attempting to create a filesystem if there is nothing there.

Nick-Rhodes-QFAB:
I will check these options - the usual approach is to go with the system default.
I can tell you that the host  I'm working on at  the moment (Galaxy-db) has these values:
sudo blockdev --getbsz /dev/vda1 
4096
sudo blockdev --getbsz /dev/vdb1 
512
If we have many small files then the 512 is appropriate and  4096 would lead to a lot of wasted space.
I will have a think and get back to you.

Nick-Rhodes-QFAB:
512 is the default, I suppose having 4096 for the root disk would make things a bit quicker - and you pretty much know what's going on there.

PR #13, opened 2020-06-17
Added slurm to dev-queue

Slugger70:


.gitignore
```diff
@@ -5,6 +5,8 @@ venv
 roles/galaxyproject.galaxy/
 roles/galaxyproject.nginx/
 roles/galaxyproject.postgresql/
+# roles/galaxyproject.slurm/ # Commented out as we have made modifications to this role so slurmdbd installation works
```
cat-bro, line 8: Could the modifications be flagged in the slurm role so that we can see where it is different from galaxyproject.slurm?  Something like `# modified for usegalaxy-au/infrastructure`

hosts
```diff
@@ -1,8 +1,16 @@
...
+dev-queue ansible_ssh_host=115.146.85.163
+
+[slurmcontrollers]
+dev-queue ansible_ssh_host=115.146.85.163
+
+[dev-slurm-head]
+dev-queue ansible_ssh_host=115.146.85.163
+
+[dev-workers]
+dev-w1 ansible_ssh_host=115.146.87.253
```
cat-bro, line 11: Where does this IP address come from?

roles/mariadb/tasks/install_redhat.yml
```diff
@@ -0,0 +1,160 @@
...
+    name: "mariadb-10.2"
+    description: "MariaDB 10.2"
+    baseurl: "http://yum.mariadb.org/10.2/centos$releasever-amd64"
+    gpgkey: "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-MariaDB"
+    gpgcheck: yes
+    enabled: yes
+
+- name: Install MariaDB yum key
+  copy:
+    content: |
```
cat-bro, line 14: Does this key need to be encrypted?

Slugger70: This is the public key for the repo

.gitignore
Slugger70: There are a lot of changes. I can put in comments.

hosts
Slugger70: Fake address.. for testing

PR #15, opened 2020-06-30
host file and variable path updates

cat-bro:
- As per Simon's work with rabbit, have `hostname: "{{ ansible_hostname }}.usegalaxy.org.au"` in all.yml
- Replace hyphens in group names with underscores.  Ansible gives a warning  `Invalid characters were found in group names but not replaced, use -vvvv to see details'.  It probably doesn't matter but it's difficult to tell.
- add vars_files to dev_playbook

### DIFF
--- a/dev_playbook.yml
+++ b/dev_playbook.yml
@@ -1,5 +1,9 @@
-- hosts: galaxyservers
+- hosts: dev_galaxy_server
   become: true
+  vars_files:
+      - group_vars/all.yml
+      - group_vars/galaxyservers.yml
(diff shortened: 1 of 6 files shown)


cat-bro:
From what I have read on the ansible github, '-' is a "bad" character in a group name  but it's not entirely clear why - I think maybe because it cannot be a python variable name.  It won't stop ansible from running.  The warning can be switched off with `TRANSFORM_INVALID_GROUP_CHARS` setting in ansible.cfg but it seems easier to adopt their convention.

PR #21, opened 2020-08-19
put admin email addresses in vault, add tasks to create users on machines

cat-bro:
these are in the vault as one variable per email address: simon_email, nick_email etc


cat-bro:
To check the output of this: `ansible-playbook -i test_hosts test_playbook.yml`

test_hosts
```
[test]
localhost ansible_connection=local
```

test_playbook.yml
```
- hosts: test
  vars_files:
      - group_vars/all.yml
      - group_vars/VAULT
      - group_vars/galaxyservers.yml
      - host_vars/dev.usegalaxy.org.au.yml
  pre_tasks:
  - name: 'print stuff'
    debug:
      msg: "{{ galaxy_admin_users }}"
```

cat-bro:
@Slugger70 I've added some code to the common role for setting up the users.  There is no `when` clause for this - the users will be set up on every machine.  Maybe there should be a `when`.

We are missing JL and NG in the machine_users list.

I'm wondering whether it would be sound to encrypt the public keys as well since they contain email addresses.  And help@genome... it has been getting a lot of spam lately.

I've tested the user creation on a different machine, this hasn't been run on dev yet.

Slugger70 (approved):

PR #30, opened 2020-10-13
Update staging terraform, add security groups

cat-bro:
Non-default rules for custom security groups have been imported and the security group configs are in dev-secgroups.tf and staging-secgroups.tf.

Due to changes in naming of images, the dev images now correspond to slightly different names in nectar ("NeCTAR Ubuntu 20.04 LTS (Focal) amd64 [v3]" instead of "NeCTAR Ubuntu 20.04 LTS (Focal) amd64".  We should probably store the image_id instead of image_name in our terraform configuration.

The is also `terraform/one-offs/rabbitmq` for the rabbitmq security group.  As a one-off, this does not need backends.tf, it's just for adding rabbitmq to a tenancy that does not have it yet.

### DIFF
--- a/terraform/dev/dev-initial.tf
+++ b/terraform/dev/dev-initial.tf
@@ -1,47 +1,47 @@
 # dev database
 resource "openstack_compute_instance_v2" "dev-db" {
   name            = "dev-db"
-  image_name      = "NeCTAR Ubuntu 20.04 LTS (Focal) amd64"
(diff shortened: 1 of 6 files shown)


cat-bro:
@Slugger70 there was an error when trying to create a c3.xlarge or m3.xlarge machine for the staging worker.  m3.large spawns ok.

Slugger70 (approved):

PR #37, opened 2020-10-15
Added staging workers and minor mod to dev worker playbook

Slugger70:


cat-bro (approved):
We could refactor this to have group_vars/workers.yml with `dev`/`staging` as a variable

Slugger70:
True.. I might make it an issue

PR #39, opened 2020-10-18
set pulsar_pycurl_ssl_library to openssl

cat-bro:
On dev pulsar jobs were not running because 'import pycurl' threw an exception.  Using this variable sets the pycurl backend correctly (the default from the role is gnutls for Debian)


Slugger70 (approved):

Slugger70:
We'll have to run this on staging also

PR #41, opened 2020-11-02
Dynamic job rules for dev and staging

cat-bro:
Modules with functions that call galaxy's `map_tool_to_destination` using a tool_destinations.yml file.  These are equivalent to using dynamic_dtd with the tool_destionations.yml file but can be extended to do things like limiting resources to a group or choosing destinations based on queue sizes while still defaulting to the destination chosen by `map_tool_to_destination`.

This has the [WIP] label - it still needs to be tested on staging.


Slugger70:
I like it. This is a really good solution. Especially the expandability. It's a nice hybrid of the tool_dests and expansion.

Slugger70 (dismissed):

Slugger70 (approved):

PR #55, opened 2020-11-10
Initial scripts for pawsey-nfs servers

Slugger70:
These need a bit of re-factoring to make them a bit more generic but this will do for now.

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -39,9 +39,9 @@ machine_users:
 galaxy_user:
   name: galaxy
   shell: /bin/bash
-  uid: 998
+  uid: 1024
   group: galaxy
-  gid: 998
+  gid: 1024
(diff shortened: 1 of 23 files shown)


group_vars/all.yml
```diff
@@ -39,9 +39,9 @@ machine_users:
 galaxy_user:
   name: galaxy
   shell: /bin/bash
-  uid: 998
+  uid: 1024
```
cat-bro, line 42: If we have uid: "{{ galaxy_user_id }}"  we can be defining this in new group_vars files for pawsey, dev, staging

Slugger70: yeah, good idea. I'll get onto it after I've stood up the workers and the queue server

cat-bro (approved):

PR #58, opened 2020-11-11
comment out enable_tool_recommendations

cat-bro:
If `enable_tool_recommendations` is true, galaxy detects extra dependencies that need to be installed but these cannot be installed.   It's not clear whether this is a bug in ansible-galaxy, galaxy or our setup.  Leave out this config option for now.


cat-bro:
https://github.com/galaxyproject/galaxy/blob/release_20.05/lib/galaxy/dependencies/__init__.py#L204

Slugger70 (approved):

PR #69, opened 2020-11-20
Add package for tools that need libGL

cat-bro:
Add libgl1-mesa-glx  for Bandage. Also fix galaxyservers.yml (this was broken for dev and staging)


group_vars/all.yml
```diff
@@ -64,6 +64,7 @@ common_packages:
     - apt-transport-https
     - tar
     - bzip2
+    - libgl1-mesa-glx
```
Slugger70, line 67: This is ok here as it's pretty innocuous. But with some of the others we need to be fixing conda upstream.

Slugger70 (approved):

PR #85, opened 2020-12-02
add 'sendfile on' to nginx conf

cat-bro:
based on https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/roles/geerlingguy.nginx/templates/nginx.conf.j2, this is probably what it needs

### DIFF
--- a/templates/nginx/galaxy.j2
+++ b/templates/nginx/galaxy.j2
@@ -47,6 +47,11 @@ server {
         include uwsgi_params;
     }
 
+    location /_x_accel_redirect/ {
+        internal;
+        alias /;
+    }
+
     location /robots.txt {
         alias {{ galaxy_server_dir }}/static/robots.txt;
     }
--- a/templates/nginx/galaxy.j2
+++ b/templates/nginx/galaxy.j2
@@ -7,6 +7,7 @@ server {
     # Our log files will go here.
     access_log  /var/log/nginx/access.log;
     error_log   /var/log/nginx/error.log;
+    sendfile    on;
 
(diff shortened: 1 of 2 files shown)


templates/nginx/galaxy.j2
```diff
@@ -7,6 +7,7 @@ server {
     # Our log files will go here.
     access_log  /var/log/nginx/access.log;
     error_log   /var/log/nginx/error.log;
+    sendfile    on;
```
Slugger70, line 10: Do we have the _x_accel_redirect stuff already in the nginx conf?

Slugger70: https://docs.galaxyproject.org/en/master/admin/nginx.html?highlight=jbrowse#sending-files-with-nginx

Slugger70: And the galaxy.yml setting also

cat-bro: we have this in our group_vars/galaxyservers.yml but it's commented out some other options I was not sure if we were using 
    # nginx_x_accel_redirect_base: /_x_accel_redirect

we also had the block in nginx configs on galaxy qld

location /_x_accel_redirect/ {
    internal;
    alias /;
}

cat-bro: I'll add these in.  Are these instead of of in addition to using 'sendfile on'?

Slugger70: Maybe, lets leave both and see what happens

Slugger70 (approved):

PR #105, opened 2020-12-15
Remove cloud configured hostname settings

Slugger70:


Slugger70:
Could we do this in a more targeted fashion? Like just make it for Pawsey instances? I think pulsar-paw will suffer from the same fate

cat-bro:
Looks like it would be safe to run on all machines

PR #120, opened 2021-02-02
Added a 9slot_1off destination to the job_conf

Slugger70:
This is for jobs that will use a lot of disk space. We can assign them 9 slots but ensure that only one of them runs at a time.


Slugger70:
Also added some commands to the pawsey-workers playbook to move the /tmp to the vdc.

cat-bro (approved):

PR #121, opened 2021-02-02
Modded internal/external ip to workaround internal ip issues with que…

Slugger70:
…ue vs galaxy


Slugger70:
This is required as pawsey queue can't talk to pawsey on internal networks. And as all slurm.confs have to be the same, the best way to handle this is with changing to external ips for queue and pawsey.

cat-bro:
Do the workers need to be added to the special group?

PR #129, opened 2021-02-04
terraform pawsey updates

cat-bro:
The backends file at pawsey was lost so terraform has been unavailable for a while.

All instances have been imported with `terraform import`.

This file needs some rearranging but I'm setting it upstream now so that it matches with the terraform.tf I just emailed to @Slugger70


cat-bro:
Simon I've just realised that I forgot to set sensible ids for the special group rules.  I can update this and send another state file.

Slugger70:
no worries. We will then copy it to the object store and reference it in backends.

Slugger70:
Let me know when this is ready to go.. We need to find somewhere to put it.

cat-bro:
This is ready to go.  I just emailed the terraform.tsfstate that goes with it.  We can't really use it until we have backends setup for it but this is an accurate reflection of the machines we have at the moment.

I have a decent workflow for detecting and importing new things we create in the meantime.

PR #144, opened 2021-02-25
Updated mothur.cluster.split destination to slurm_9slots

Slugger70:


cat-bro (approved):
the bad job crashed on slurm2 and slurm16.  It's probably the same inputs

cat-bro (approved):

PR #152, opened 2021-03-02
Removed legacy (Brisbane) RabbitMQ server from Terraform

Slugger70:


Slugger70:
This Terraform was for the old RabbitMQ server that connected pulsar-mel2 and pulsar-mel3 to the Brisbane Galaxy server. We do not need this here anymore.

cat-bro (approved):

Issue #156, opened 2021-03-03
Upgrade to latest galaxy role version


cat-bro:
At present this is `0.9.11`

PR #169, opened 2021-03-08
upgrade galaxyproject roles to current versions

cat-bro:
add `data_dir: "{{ galaxy_root }}/cache"` because otherwise all of the cache directories spawn within `/mnt/galaxy`

Addresses #156

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,9 +1,9 @@
 - name: galaxyproject.galaxy
-  version: 0.9.6
+  version: 0.9.11
 - name: galaxyproject.nginx
   version: 0.6.4
 - name: galaxyproject.postgresql
-  version: 1.0.2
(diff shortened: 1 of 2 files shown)


requirements.yml
```diff
@@ -20,7 +20,7 @@
 #- src: galaxyproject.slurm - Waiting for a new release that sets up clustering for dbd properly
```
Slugger70, line 20: Is there an updated slurm role too?

cat-bro: yes, 0.1.3 released 2 months ago but it looks like the only new commit is this one https://github.com/galaxyproject/ansible-slurm/commit/a3104b029f02c161cc983f0f30551e8d516d3377

Slugger70 (approved):

PR #180, opened 2021-03-16
increase nginx client_max_body_size to 20g

cat-bro:


Nick-Rhodes-QFAB:
Let's see how we go with 20G!

PR #202, opened 2021-03-29
Added pulsar-mel2

Slugger70:


cat-bro (approved):
Does the worker playbook need the block that mounts the volume?

Slugger70:
yes, they have 100gb volumes for tmp and cvmfs.

PR #210, opened 2021-04-06
pulsar-post-tasks as a role

cat-bro:
change the perms from 750 to 754 on pulsar roles so that regular users can read files in working dirs.  Group these into a role because all pulsars need the same post tasks


roles/pulsar-post-tasks/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Install slurm-drmaa
+  package:
+    name: slurm-drmaa1
+    state: present
+- name: own the /mnt/pulsar dirs as ubuntu
+  file:
+      path: /mnt/pulsar
```
Slugger70, line 8: Change these to the variable for pulsar root...

roles/pulsar-post-tasks/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+      owner: ubuntu
+      group: ubuntu
+      state: directory
+      recurse: yes
+      mode: 0754
+- name: own the /mnt/galaxy dirs as ubuntu
```
Slugger70, line 14: same here

Slugger70 (changes requested):

roles/pulsar-post-tasks/tasks/main.yml
cat-bro: good thinking

roles/pulsar-post-tasks/tasks/main.yml
cat-bro: I've added this as `galaxy_dir` in pulsarservers.yml.  I didn't want it to clash with galaxy_root from group_vars/galaxyservers.yml even though it is not in scope here

cat-bro:
I put perms up to 755 for /mnt/pulsar so users can cd.  Is this getting too permissive?  An alternative would be adding ubuntu to the devs group in the common role, then the machine users would have r and x.

Slugger70 (approved):

PR #226, opened 2021-04-21
Absorb volume setup into common role

cat-bro:
instead of the pasted block in playbooks, have a variable

```
attached_volumes:
  - fstype: ext4
    device: /dev/vdd
    path: /hello/world
```

then the common role goes through the block of creating the fs and mounting the volume prior to doing anything else.

This works for dev and pulsar-nci-test, will refactor the others if this logic seems sensible


Slugger70 (approved):

Slugger70:
Nice!

PR #241, opened 2021-05-06
playbook and script for updating the tool labels on galaxy

cat-bro:
This can be its own jenkins job triggered at the end of the install and update jobs.  The script is copied to galaxy and run as the galaxy user in a working directory, then the altered shed_tool_conf.xml is copied to the shed_tool_conf_path.  There is an integrity check so that if shed_tool_conf.xml has been altered by actual galaxy while the script has been running, the file will not be replaced.


Nick-Rhodes-QFAB:
Looks good to me.

PR #242, opened 2021-05-06
send minimap to pulsar-mel2 and slurm9

cat-bro:
Due to queues for pulsar-mel3_big..

Up to 500Mb to pulsar-mel_mid, 500Mb to 5G to pulsar-mel_big, default dest slurm_9slots


cat-bro:
it really depends what day it is whether this is a good choice.. next week maybe it will be slurm cluster that is overloaded

PR #248, opened 2021-05-14
[wip] Remote pulsar cron role

cat-bro:
Daily cron jobs run as ubuntu user to remove old job working directories on pulsar with a set of parameters.

There are almost certainly some syntax errors in the resulting python so I'll set this as a draft, template some files and try them out.


Slugger70:
Can you add a .j2 to the end of the filename for the template please? That way we'll know it's a template and not a useable python script.

Slugger70 (approved):

PR #260, opened 2021-05-28
Updates for dev

cat-bro:
Get rid of custom security groups and use default group, move pulsars from aarnet dev back to dev, use latest rabbit version for dev only


cat-bro:
I need to update the dev-db password because it has a % in it and it's messing with my attempts to fix the cron psql thing

Slugger70 (approved):

PR #281, opened 2021-06-22
delete-tmp-jwds-cron role

cat-bro:
this is on pawsey in /home/ubuntu/deleting_things.  The python script writes a list of rm -rf commands for job working directories for error jobs.  It's not set up to run the bash script automatically yet, just trying it out


cat-bro:
this should not run on the head node but instead be copied over to pawsey-user-nfs

cat-bro:
This script has been run on pawsey for the last year to get rid of old working directories.  It would be better to copy the script over to pawsey-job-nfs and run it there, but there does not seem to be any access to pawsey-job-nfs through external or internal IPs.

Slugger70 (approved):

Issue #97, opened 2020-12-04
read-only role does not get permission to see tables


cat-bro:
The read-only user has correct access on dev and staging but not on pawsey.  The ubuntu user (using galaxy role) has correct access on pawsey.

UPDATE: If the postgres-objects role is run when the db is empty the read-only user is not granted any table permissions.  Not sure how to fix this apart from rerunning the db role after the first run of the galaxy role.

cat-bro:
The workaround we now have in place for this is to set the value for `postgresql_objects_privileges` to a different value the first time the playbook is run (detected by the existence of the production data directory). This is a shaky solution as it depends on the playbook running smoothly the first time but it is probably good enough.

PR #293, opened 2021-07-02
move reader and tiaas db user creation to galaxy playbook on dev

cat-bro:
This is a draft. It has not been tested on dev and dev-db

Would solve #97 and errors running the db playbook the first time


cat-bro:
this does not work

PR #300, opened 2021-07-12
Added file sources config

nuwang:


Slugger70:
Awesome. Is this compatible with our current installation or does it require your changes to 21.05? As this requires the file sources config file directive in Galaxy we might not enable it on prod or staging unril it's tested on dev.

In galaxy.yml:
```
file_sources_config_file
Description
Configured FileSource plugins. The value of this option will be resolved with respect to <config_dir>.

Default
file_sources_conf.yml

Type
str
```

If it's automatically configured just by having a non-sample file_sources_config.xml file present then we may want to just do it on the dev server first.

nuwang:
I rebased that branch over the weekend so it's ready to test: https://github.com/usegalaxy-au/galaxy/tree/release_21.05_cloudstor

Good point about testing on dev, will move it to dev configs first.

nuwang:
Have made the change. Reverting this commit: https://github.com/usegalaxy-au/infrastructure/pull/300/commits/0c088cc47938f9514c547812c63d9180614912e9 should allow us to restore production settings.

Slugger70:
Ahh. Awesome

Slugger70 (approved):

PR #324, opened 2021-07-29
set destinations for meryl and merqury

cat-bro:
meryl can run on pulsar.

eu sets mem at 12 for merqury and 110 for meryl.


cat-bro:
The other VGP tools we installed are not listed in eu's tool_destinations doc

Slugger70 (approved):

PR #345, opened 2021-08-12
upgrade tiaas role

cat-bro:
It asks for fewer select permissions for the tiaas db user and an additional delete permission for group_role_association.  There is also `tiaas_disassociate_training_roles`:

```
# Create a cron job to disassociate training roles from groups after trainings have expired, set to `false` to disable
tiaas_disassociate_training_roles:
  hour: 0      # optional, defaults to 0
  minute: 0    # optional, defaults to 0
```

Looks like sets a daily cron job to delete sessions that have finished at an hour and minute that we can set, or have `tiaas_disassociate_training_roles: false` to not set the cron.

I've tentatively set these to `hour: 9` so that they clean up at 7pm

### DIFF
--- a/host_vars/aarnet-db.usegalaxy.org.au.yml
+++ b/host_vars/aarnet-db.usegalaxy.org.au.yml
@@ -73,7 +73,7 @@ postgresql_objects_privileges:
     privs: DELETE
   - database: galaxy
     roles: tiaas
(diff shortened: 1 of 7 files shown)


Slugger70 (approved):

cat-bro:
For reference, this is the diff between 0.08 and 0.01 https://github.com/usegalaxy-eu/ansible-tiaas2/compare/0.0.1...0.0.8

PR #351, opened 2021-08-17
Backup ansible

Slugger70:


cat-bro:
Since this is build as well as backup it could install ansible 2.10 and clone the infrastructure repo

cat-bro (approved):

Slugger70:
Ok, I'll add all that when I do the next bit - adding in the history mailer and the user grouper.

PR #378, opened 2021-08-30
pawsey job conf and tool destination changes

cat-bro:
Baseline changes to dtd, there will be another commit to take out mel3 for the time being


cat-bro:
I also copied sanitize_whitelist.txt to sanitize_allowlist.txt on pawsey so there is no need for sanitize_whitelist in host_vars

Slugger70 (approved):

PR #391, opened 2021-09-03
add another worker to mel3

cat-bro:
the image names of the existing VMs needed to be changed to match their image ids, otherwise terraform wanted to replace them

### DIFF
--- a/group_vars/pulsar_mel3/pulsar-mel3_slurm.yml
+++ b/group_vars/pulsar_mel3/pulsar-mel3_slurm.yml
@@ -44,7 +44,7 @@ slurm_nodes:
 
 slurm_partitions:
     - name: main
-      nodes: "pulsar-mel3-w1,pulsar-mel3-w2,pulsar-mel3-w3,pulsar-mel3-w4,,pulsar-mel3-w6"
+      nodes: "pulsar-mel3-w1,pulsar-mel3-w2,pulsar-mel3-w3,pulsar-mel3-w4,pulsar-mel3-w5,pulsar-mel3-w6"
       Default: YES
       MaxTime: INFINITE
       State: UP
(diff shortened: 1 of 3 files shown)


group_vars/pulsar_mel3/pulsar-mel3_slurm.yml
```diff
@@ -36,10 +36,15 @@ slurm_nodes:
...
+    - name: pulsar-mel3-w6
+      NodeAddr: "{{ hostvars['pulsar-mel3-w6']['ansible_ssh_host'] }}"
+      CPUs: 16
+      RealMemory: 64419
+      State: UNKNOWN
 
 slurm_partitions:
     - name: main
-      nodes: "pulsar-mel3-w1,pulsar-mel3-w2,pulsar-mel3-w3,pulsar-mel3-w4,pulsar-mel3-w5"
+      nodes: "pulsar-mel3-w1,pulsar-mel3-w2,pulsar-mel3-w3,pulsar-mel3-w4,,pulsar-mel3-w6"
```
Slugger70, line 47: Missing w5 here

terraform/pulsar-mel3/pulsar-mel3-initial.tf
```diff
@@ -1,7 +1,7 @@
 # Pulsar-mel3
 resource "openstack_compute_instance_v2" "pulsar-mel3" {
   name            = "pulsar-mel3"
-  image_name      = "NeCTAR Ubuntu 20.04 LTS (Focal) amd64"
+  image_name      = "NeCTAR Ubuntu 20.04 LTS (Focal) amd64 [v9]"
```
Slugger70, line 4: It may be better to go with the image id here. But the versioned name is probably ok too

Slugger70 (changes requested):

terraform/pulsar-mel3/pulsar-mel3-initial.tf
cat-bro: agreed, image id is better because it won't go out of date

Slugger70 (approved):

PR #395, opened 2021-09-06
Moved hifiasm to hm2 and added gxadmin to .gitignore

Slugger70:


Slugger70 (commented):
Going to admin merge this.. :)

PR #400, opened 2021-09-09
Added rules for abyss, shasta and raven

Slugger70:


Nick-Rhodes-QFAB:
File sizes to infinity (and beyond!)

PR #409, opened 2021-09-14
Nci training initial setup

Slugger70:


cat-bro (approved):
Lovely!  I really like the loop definition for the workers in terraform

PR #433, opened 2021-09-22
send pulsar-mel_big jobs to mel3 temporarily

cat-bro:
there are a bunch of jobs for pulsar-mel_big that will not queue because pulsar failures last night have caused that destination to become full while nothing is running.  Temporarily send these jobs to pulsar-mel3_mid which is the same size.

with total perspective vortex we will never have to worry about this happening..


cat-bro:
actually this might not be necessary.  Somebody deleted their queued jobs for pulsar-mel_big, so jobs have started running there again

PR #440, opened 2021-09-29
add query for jobs running at a given datetime

cat-bro:
Local gxadmin query to find jobs created before and finished after a given datetime, to find jobs that might have two sets of dataset files after the object store changes.


cat-bro:
also fixing some formatting problems with the previous query

Slugger70 (approved):

PR #448, opened 2021-10-04
method for adding internal ssh keys

cat-bro:
A set of tasks in the common role `extra_keys.yml` for adding internal ssh keys to machines

The internal ssh keys are defined in all.yml:
```
internal_ssh_keys:
  ubuntu_maintenance_key:
    private_key_content: "{{ vault_ubuntu_maintenance_key }}"
    public_key_src: files/keys/ubuntu_maintenance_key.pub
    user: ubuntu
```

If any machine needs to have the private or public part of this key it is in the host_vars, i.e.
```
extra_keys:
  - id: ubuntu_maintenance_key
    type: public
    from: "{{ hostvars['dev']['ansible_ssh_host'] }}"
```
or
```
extra_keys:
  - id: ubuntu_maintenance_key
    type: private
```

The tasks forbid listing a public key in extra_keys without a `from`, so that access is as restricted as possible.


Slugger70 (approved):
Good work. I like the implementation

PR #480, opened 2021-10-21
Attempt at adding disk partitioning to volumes

Slugger70:


host_vars/qld-pulsar-himem-0.yml
```diff
@@ -18,7 +18,8 @@ pulsar_rabbit_username: "galaxy_qld_hm0"
 pulsar_rabbit_vhost: "/pulsar/galaxy_qld_hm0"
 
 attached_volumes:
-  - device: /dev/vdb1
+  - device: /dev/vdb
+    partition: 1
```
Slugger70, line 22: this is an optional var and we can use it on all new VMs and this role should still work with old ones..

roles/attached-volumes/tasks/set_up_volume.yml
```diff
@@ -5,11 +5,22 @@
         device: "{{ attached_volume.device }}"
         unit: GiB
       register: volume_info
-    - name: Create a filesystem on volume
+    - name: create a partition on volume
+      parted:
+        device: "{{ attached_volume.device }}"
+        number: "{{ attached_volume.partition }}"
+        state: present
+      when: volume_info.disk is defined and volume_info.disk.dev is defined and volume_info.disk.dev == attached_volume.device and partition is defined
```
cat-bro, line 13: `attached_volume.partition`, also on lines 22 and 23

PR #483, opened 2021-10-22
Create role login-override

neoformit:
### A new role which applies a short-term implementation for a custom GA login page

- Check if file `Login.vue` has been updated in Galaxy core
- If not, replace with our custom `Login.vue` and rebuild client

**To do**
- [x] Fail if `Login.vue` checksum doesn't match (require manual check and update checksum)
- [x] Activate venv for client build?
- [x] Check client build command
- [x] Is `galaxy_login_checksum_expected` ok in root of `galaxyservers.yml`?
- [ ] Test whether `Login.vue` actually works

### DIFF
--- a/dev_playbook.yml
+++ b/dev_playbook.yml
@@ -46,9 +46,7 @@
     - usegalaxy_eu.tiaas2
     - geerlingguy.docker
     - usegalaxy_eu.gie_proxy
-    - role: login-override
-      become: true
-      become_user: galaxy
+    - login-override
   post_tasks:
     - name: Clone Galaxy-Owncloud-Integration
       git:


group_vars/galaxyservers.yml
```diff
@@ -1,3 +1,6 @@
+# Login override
+galaxy_login_checksum_expected: 4ec5fb5a6f66fdd8dce0166aafceb0af
```
cat-bro, line 2: could this go in `host_vars/dev.usegalaxy.org.au.yml`?

roles/login-override/tasks/login_override.yml
```diff
@@ -0,0 +1,8 @@
+---
+  - name: Overwrite Galaxy core Login.vue
+    copy:
+      src: {{ galaxy_config_file_src_dir }}/client/Login.vue
+      dest: {{ galaxy_root }}/client/src/components/login/Login.vue
```
cat-bro, line 5: galaxy_server_dir instead of galaxy_root

roles/login-override/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+  - name: Get checksum for Galaxy Client Login.vue
+    shell: "md5sum {{ galaxy_root }}/client/src/components/login/Login.vue | awk '{print $1;}'"
+    register: galaxy_login_checksum
```
nuwang, line 4: We could alternatively use the ansible patch module: https://docs.ansible.com/ansible/latest/collections/ansible/posix/patch_module.html and apply a git diff. If the patch fails to apply cleanly, it will be detectable as an incompatible change. As long as we know which version the patch last applied cleanly to, it should also be possible to see the code in full context if necessary? This is assuming the changes are fairly incremental. If the changes are substantive, this checksum approach makes more sense?

PR #486, opened 2021-10-24
New stats server

Slugger70:


Slugger70:
I still need to add the new stats role but atleast the stats server is here.

cat-bro (approved):

PR #495, opened 2021-10-26
tpv production rules

cat-bro:
Conversion of tool_destinations.yml to new rules.  There are some #todos in the script, the script needs to convert tool ids, there needs to be a default tool and rules for destinations and roles.


cat-bro:
Additions needed for tools/destination *_auto.yml for production:

add default, pulsar_preferred from the dev vortex_config (debate pulsar_preferred vs explicit schedule key for all tools that go to pulsar)
pulsar-paw destination requires pulsar-paw tag
pulsar-nci-training destination accepts training tag
slurm accepts training tag
pulsar-paw compatible tools accept pulsar-paw tag
Assume for now that interactive_pulsar tools are still handled from the job conf, adjust script to not include this

cat-bro:
The tools that will go to pulsar currently have `prefer: pulsar` but they should have `accept: pulsar` instead, otherwise they will never go to slurm regardless of free space.

nuwang:
> The tools that will go to pulsar currently have `prefer: pulsar` but they should have `accept: pulsar` instead, otherwise they will never go to slurm regardless of free space.

Is that a bug? Shouldn't it go to slurm once we have the rank function hooked in?

cat-bro:
It's not a bug. The current ranking function has

```
candidate_destinations.sort(key=lambda d: (-1 * d.score(resource), cpu_by_destination.get(d.id)))
```
 If it sorts by score before availability, a preferred destination will always be returned at the front of the list even if an accepted destination has higher availability. The sorting could be updated.

host_vars/pawsey.usegalaxy.org.au.yml
```diff
@@ -130,7 +138,11 @@ host_galaxy_config:  # renamed from __galaxy_config
     nginx_upload_job_files_store: "{{ nginx_upload_job_files_store_dir }}"
     nginx_upload_job_files_path: '/_job_files'
     job_config_file: "{{ galaxy_config_dir }}/job_conf.yml"
+<<<<<<< Updated upstream
     dependency_resolvers_config_file: "{{ galaxy_server_dir }}/config/dependency_resolvers_conf.xml.sample"  # remove for 21.09
+=======
+    watch_job_rules: true  # important for total perspective vortex
```
nuwang, line 144: Looks like a merge conflict here

files/galaxy/dynamic_job_rules/pawsey/total_perspective_vortex/default_tool.yml.j2
```diff
@@ -0,0 +1,52 @@
...
+        'pulsar-qld-himem-2': 'pulsar-qld-high-mem2',
+      }
+      params = {
+        'pretty': 'true',
+        'db': 'queues',
+        'q': 'SELECT last("percent_allocated") from "sinfo" group by "host"'
+      }
+      if len(candidate_destinations) > 1:
+        try:
+          response = requests.get('https://stats.usegalaxy.org.au:8086/query', auth=('grafana', '{{ vault_influx_grafana_password }}'), params=params)
```
nuwang, line 37: Maybe we could read the `vault_influx_grafana_password` off os.environ?

cat-bro: the way that the galaxy user is set up, it has its own .bashrc file so it could have export INFLUX_GRAFANA_PASSWORD=xyz

Slugger70 (approved):

PR #513, opened 2021-11-09
send abricate to pulsar-mel2

cat-bro:
the abricate env is broken on pulsar-mel3, it can't be rebuilt at the moment because of bioconda


cat-bro:
admin-merging this because of user tickets

PR #576, opened 2021-12-01
send large hisat2 jobs to high mem

cat-bro:
send hisat2 jobs with inputs over 20GB to high mem pulsars.


Nick-Rhodes-QFAB:
makes good sense!

PR #577, opened 2021-12-02
reduce mem per core for tools that go to pulsar-paw

cat-bro:
8-core jobs on pulsar-paw are failing.  The native specification is asking for more memory than the 30500MB allowed per node.  This value could probably be higher in the slurm config but for now this is a quicker fix.


cat-bro:
This predates the switchover to total perspective vortex.  The memory in /proc/meminfo is almost exactly the same as that for the pulsar-mel2 machines

Nick-Rhodes-QFAB:
Merging after discussion

PR #581, opened 2021-12-07
More azure

Slugger70:


roles/pulsar-post-tasks/tasks/main.yml
```diff
@@ -3,6 +3,7 @@
   package:
     name: slurm-drmaa1
     state: present
+  when: install_drmaa is defined and install_drmaa
```
cat-bro, line 6: adding this will prevent drmaa from being installed on pulsars without `install_drmaa: true` somewhere.  Could be in group_vars/pulsarservers.yml

cat-bro:
great fix btw, it was lovely when jobs started running on the azure machine with the new compilation of slurm-drmaa

roles/pulsar-post-tasks/tasks/main.yml
Slugger70: good idea

cat-bro (approved):

PR #609, opened 2022-01-09
default tool: only map destination ids when they are different from stats api + set pulsar-high-mem2 to offline

cat-bro:
The default tool code uses the dict 'grafana_destinations' to map each destination id to the id that will come from the stats api. This is usually the same as the destination id. Because pulsar-qld-blast was not added to this dict there was an error in the ranking function. This change means it will default to the destination id if the key is not in the 'grafana_destinations' dict and new destinations can be added to destinations.yml without breaking anything.


cat-bro:
I altered default_tool.yml on pawsey on xmas eve because the ranking function was defaulting to weighted random sampling with a KeyError for 'pulsar-qld-blast'

Slugger70 (approved):

PR #621, opened 2022-01-19
reduce mem multiplier for blast pulsar jobs

cat-bro:
using 3.3 means that each job gets slightly more than a quarter of the RAM and only three jobs can run at a time.  Reduce it to 3.1 so that a 4th job can fit.


cat-bro:
I remember us calculating that in the office.  I think the discrepancy is because we didn't factor the difference between 1000 and 1024

Slugger70 (approved):

PR #644, opened 2022-02-16
use interactive_tools.yml in tpv

cat-bro:
2 ITs are disabled (jupyter and rstudio).  Having interactive tools go through tpv would make it possible to reenable them for some users.

Also added a toolbox filter to hide the tools from users that cannot access them.   Both the filter and TPV only allow the tools for users with the trusted_for_ITs role but this could be extended to include the training-smorgasbord role or all training roles. [Updated: this now includes all training roles]

rstudio and jupyter are uncommented in the tool_conf_interactive.


cat-bro:
this will need more rebasing and refactoring than is worthwhile so can be closed.

PR #645, opened 2022-02-17
use real path to links.xml

TomHarrop:
Singularity doesn't seem to like the symlink

```bash
FATAL:   container creation failed: mount /mnt/galaxy/local_tools/links->/mnt/galaxy/local_tools/links error: while mounting /mnt/galaxy/local_tools/links: mount source /mnt/galaxy/local_tools/links doesn't exist
```


cat-bro (approved):

Slugger70:
Singularity won't use symlinks in bind mounts. Same as docker.

PR #648, opened 2022-02-18
send cat_prepare to high mem pulsar

cat-bro:


cat-bro:
on second thoughts this tool should not be on GA.  We need to use the data manager data_manager_cat and send the job to the 32 core slurm node.

PR #660, opened 2022-02-24
Change custom indices path on pulsars and in galaxyservers.yml

cat-bro:
- one-off playbook to move custom-indices down one level on pulsars
- mount custom-indices instead of tools-indices on pulsar workers
- galaxy_custom_indices_dir in galaxyservers for each galaxy host_vars file

staging and dev are not affected because they do not have any ref data

This depends on #658 being merged.
When switching over, before running the pawsey playbook I will need to run `find /mnt/galaxy/tools-indices/tool-data -type f -exec sed -i -e 's/tools-indices\/custom-indices/custom-indices/g' {} \;` to update the loc files


cat-bro:
Marking this as ready to review (but still do not merge)

Slugger70 (approved):

PR #665, opened 2022-03-01
Install cactus 

TomHarrop:
Hi @cat-bro , can I install cactus from /home/tom for testing?

I also removed LINKS since it's in toolshed / tools-iuc now - but not sure about this.


cat-bro:
Best to copy it into /mnt/galaxy/local_tools since if it runs with docker or singuarity /home/tom will be outside of the readable locations

cat-bro:
if it's conda-only /home/tom is fine

cat-bro (approved):

PR #671, opened 2022-03-01
Aarnet playbook update

cat-bro:
Some updates for aarnet and aarnet worker playbooks.  This also adds a role 'install-tpv' because it needs to run after common and before galaxyproject.galaxy.

toolbox filters are moved out of galaxyservers and into host files because they cannot be copied before galaxy code exists but must be copied before galaxy is restarted... the only quick workaround I can think of is to comment them out for the first run and remember to uncomment them after.


aarnet_playbook.yml
```diff
@@ -7,7 +7,8 @@
       - group_vars/galaxyservers.yml
       - group_vars/aarnet_slurm.yml
       - host_vars/aarnet.usegalaxy.org.au.yml
-      - host_vars/aarnet_job_conf.yml
+      - secret_group_vars/stats_server_vault # TODO: are any new variables needed for this?
```
Slugger70, line 10: No

host_vars/aarnet.usegalaxy.org.au.yml
```diff
@@ -136,6 +160,7 @@ db_address: "{{ hostvars['aarnet-db']['internal_ip'] }}"
 gxadmin_ubuntu_config_dir: /home/ubuntu/.config
 
 # vars for stats_collection
+# TODO: does this need updating?  Ask SG
 stats_dir: /home/ubuntu/stats_collection/
 db_server: "{{ hostvars['aarnet-db']['internal_ip'] }}"
 influx_url: "stats.genome.edu.au"
```
Slugger70, line 166: needs to be stats.usegalaxy.org.au

Slugger70 (changes requested):

Slugger70 (approved):

PR #691, opened 2022-03-09
bind /etc/passwd in docker

TomHarrop:
Hi @cat-bro 

This was [suggested by the developers](https://github.com/ComparativeGenomicsToolkit/cactus/issues/677#issuecomment-1063211195) as a workaround to the issue running Cactus with docker on dev.

I *think* Singularity mounts `/etc/passwd` by default but Docker doesn't.

Not sure if this is the right way to do it in TPV.

Thanks!

### DIFF
--- a/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/vortex_config.yml
+++ b/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/vortex_config.yml
@@ -203,14 +203,14 @@ tools:
     cores: 3
     params:
       docker_enabled: true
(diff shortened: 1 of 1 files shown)


cat-bro:
these are the defaults for docker volumes: https://github.com/galaxyproject/galaxy/blob/dev/test/unit/app/jobs/job_conf.sample_advanced.yml#L425

what you have looks right but it should probably be `/etc/passwd:ro`

cat-bro (approved):

cat-bro:
this is on dev now

PR #712, opened 2022-03-21
send cactus to pulsar with docker and maximum memory, add cvmfs to volumes

cat-bro:
@TomHarrop I'm wondering if this might be ok without binding /etc/exports because of `docker_set_user: 1000`


TomHarrop:
@cat-bro I'm not sure either way, but I'm happy to test after you've changed it.

PR #717, opened 2022-03-23
specify cvmfs bind

TomHarrop:
I think I need to bind `/cvmfs` for these tools because the tool uses reference genomes from the `all_fasta` table. Not sure if this is the right way.

Right now I'm getting a "file not found" error which suggests to me that the path isn't mounted when the container runs.


cat-bro (approved):
Can also use `inherits: slurm_docker` to get the params from there

PR #718, opened 2022-03-23
reduce cactus mem to run on dev

TomHarrop:


cat-bro (commented):
This needs the other docker params.  Could add docker_enabled etc or use 'inherits: slurm_docker'.  The docker volumes then need to be whatever is used for slurm, plus /etc/passwd

TomHarrop:
Yikes, what am i doing. thanks for checking!

cat-bro (approved):

PR #723, opened 2022-03-28
aarnet nfs playbooks

cat-bro:
Some updates to aarnet nfs playbooks.  For user-nfs there will be one giant volume instead of the smaller ones.  When the existing folders are copied over will they remain separate subfolders: user-data, user-data-2 etc?  I've chosen `/mnt` as the path for that volume on the machine but could change it to `mnt/data` to allow for future volumes at subdirectories of /mnt.

I've also added a 200G volume to the pulumi for misc-nfs.  Will check with aarnet tomorrow whether this will be OK.

### DIFF
--- a/aarnet-job-nfs_playbook.yml
+++ b/aarnet-job-nfs_playbook.yml
@@ -26,7 +26,6 @@
           path: "{{ galaxy_job_dir }}"
           owner: galaxy
           group: galaxy
-          recurse: yes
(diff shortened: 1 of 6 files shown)


cat-bro:
This should be good to go once we're sure of the mount points for the volumes

Slugger70 (approved):

PR #741, opened 2022-04-04
Added a separate gpu destination with separate limits

Slugger70:
I thought it might be a better idea in terms of limits to have a separate destination for gpu work than normal work on azure. That way we can control the limits in a more fine grained manner.


cat-bro:
About to add 'approve'.  The jenkins script is set up to copy TPV files but skip the step that copies the job conf and restarts.  Once this is merged, the pawsey_update_job_conf playbook needs to be run by a human so that galaxy will get the new config and restart (or before it is merged, from the branch).

cat-bro (approved):

PR #769, opened 2022-04-27
one-off task for updating conda environments

cat-bro:
Maintain a yaml list of conda environments that differ from what is built from the recipe.  This includes old versions of multiqc, old versions of bowtie and environments that depend on old bowtie. We can use ansible to iterate through the list of environments, preinstall those that do not exist and fix those that do.

### DIFF
--- a/one-offs/conda_envs/aarnet_conda_playbook.yml
+++ b/one-offs/conda_envs/aarnet_conda_playbook.yml
@@ -1,0 +1,11 @@
+- hosts: aarnet_galaxy_server
+  become: true
+  become_user: galaxy
+  vars_files:
+    - ../../group_vars/all.yml
+    - ../../group_vars/galaxyservers.yml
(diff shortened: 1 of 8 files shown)


cat-bro:
And TODO: try to push some of these changes back to bioconda, but we will need these tools to work sooner than this could be achieved.

cat-bro:
This list was compiled while fixing tools on aarnet in April 2022.  The playbook uses tasks from the ansible-miniconda playbook to install conda virtual environments with the right name but with extra packages to prevent software errors, e.g. tbb for older bowtie versions.

The playbooks will accept the extra variable `env_names` if you are only installing one environment.

There are playbooks for aarnet and the general pulsars.

I'm in favour of merging this because the yaml file with the environments could be a useful reference for other Galaxy admins.

Slugger70 (approved):

PR #770, opened 2022-04-27
try to force protk tools to use their docker container

TomHarrop:
Hi @cat-bro ! I'm not sure if this will work.

These protk tools have multiple entries in the `requirements` field, e.g.

```xml
    <requirements>
        <container type="docker">iracooke/protk-1.4.3</container>
        <requirement type="package" version="1.4.3">protk</requirement>
   </requirements>
```
**Only** the docker container works - the conda package and the biocontainer are broken.

I want to see if setting `use_singularity` in the vortex config will pull the container off docker hub or biocontainers... of if there's another way to do this without changing the wrapper

Also, one of the tools writes to its own installation directory in the container, hence the need for `'--writable-tmpfs'`

cc @mthang

### DIFF
--- a/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/vortex_config.yml
+++ b/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/vortex_config.yml
@@ -246,10 +246,7 @@ tools:
     params:
       singularity_enabled: True
       singularity_run_extra_arguments: '--writable-tmpfs'
-      singularity_volumes: '$defaults'
-      singularity_default_container_id: '/cvmfs/singularity.galaxyproject.org/all/python:3.8.3'
-
-
+      dependency_resolution: 'none'
 
 users:
   default:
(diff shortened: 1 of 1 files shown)


cat-bro:
It ought to pull it from docker-hub.  It's OK to leave out singularity_volumes and singularity_default_container_id now because every destination in the job conf a per-destination value for these.  You could also use `dependency_resolution: 'none'` if it doesn't need to use the conda requirement.

cat-bro:
PS. it will never run with singularity or docker without `singularity_enabled: true` or `docker_enabled: true`, without one of these it will only look for a conda environment

cat-bro (approved):

TomHarrop:
Great, thanks. I'm hoping it uses Singularity to run the container from docker-hub, but NOT the container from biocontainers... I'll see what happens when I test it.

cat-bro:
oh I see what you mean

PR #786, opened 2022-05-17
Production singularity

cat-bro:
Updated job conf to have destination parameters for docker and singularity.  In this setup a job destination could be pulsar-mel3 but would run with singularity only if the param `singularity_enabled: true` were set for the tool in the TPV.  The only parameter required would be `singularity_enabled`, everything else (`singularity_volumes` etc) lives on the job conf.  This has already been done on dev.

Every job will have in their destination parameters `singularity_enabled: False` or `singularity_enabled: True` and `docker_volumes` and `singularity_volumes`.  There is enough room on the destination_params field to have all of this data, and it's nothing in size compared to stdout/stderr for some jobs. 

An alternative design choice would be to have multiple destinations per cluster, i.e. `pulsar_mel3`, `pulsar_mel3_docker` and `pulsar_mel3_singularity` and use tags such as `singularity` to send jobs to the right destination.  It would be easy to map these destination names to the usage data used in the ranking function.

@nuwang @Slugger70 I'm interested in what you think of this given that GA's TPV config will be the example copy.  Would the `resubmit` example that Nuwan was talking about benefit from having destinations that are more specific?

### DIFF
--- a/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/default_tool.yml
+++ b/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/default_tool.yml
@@ -9,6 +9,7 @@ tools:
     params:
       nativeSpecification: "--nodes=1 --ntasks={cores} --ntasks-per-node={cores} --mem={round(mem*1024)}"
       submit_native_specification: "--nodes=1 --ntasks={cores} --ntasks-per-node={cores} --mem={round(mem*1024)}"
+      docker_memory: '{mem}G'
     scheduling:
       reject:
         - offline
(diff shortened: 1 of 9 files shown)


nuwang:
I like the simplicity of this. However, I do agree that if we want to support resubmissions, having multiple destinations would probably work better. Otherwise, if `singularity_enabled: True` is defined on a tool, and we want to resubmit that to a non-singularity destination, we would have to figure out a way to flip that bit. Of course, we did discuss destination level overrides, so I think if we implement that, that should also solve the issue?

cat-bro:
If the tool is resubmitted and the next destination is `vortex_dispatcher` it would go back through the loop and get appropriate parameters.  Once TPV can handle the nested `params` I'll try this out on dev and see what happens.

nuwang (approved):
This looks good to me.

Slugger70 (approved):

PR #793, opened 2022-05-20
fix alignment in users.yml

cat-bro:
only the top user is counted because of an indentation error


cat-bro:
@Nick-Rhodes-QFAB I'm trying to test pulsar-high-mem2 to see if it can be brought back online but because of my mistake, none of the user rules are working

PR #857, opened 2022-06-28
Test user jobs run with singularity enabled if history has 'singularity' tag

cat-bro:
- remove high-mem rule from dev default user (this contains errors fixed in the production rule, besides, dev has no high-mem destinations)
- add test_user with the rule that any toolshed tool will run will run on singularity if the history has a 'singularity' tag.  All test user accounts inherit this rule.

I got this idea from Galaxy Main because testers were encouraged to add a 'training' tag to their histories to have their jobs run on priority resources.  This seems like a neat way to control a job destination from the outside.  I'm not suggesting anything like this for general users, just for test user accounts that only we have access to.


Slugger70 (approved):

Slugger70:
Good idea

PR #868, opened 2022-07-04
[WIP] alphafold-dev on production

cat-bro:
Not sure where to put the alphafold.xml etc.  The galaxy-local-tools path is a clone of tools-au which is updated with each playbook run.  Ideally it would be somewhere that could be written to without sudo.


cat-bro:
superseded by #941, #942

PR #883, opened 2022-07-27
Give dev a hierarchical object store

cat-bro:
Replicate the production object store setup on the dev with the aim of migrating this to a distributed object store with new datasets being stored by UUID. The old directories would have weight 0 meaning that nothing would be written to them.

### DIFF
--- a/dev_playbook.yml
+++ b/dev_playbook.yml
@@ -52,6 +52,15 @@
     - dj-wasabi.telegraf
     - login-override
   post_tasks:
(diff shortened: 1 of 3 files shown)


cat-bro:
Replicate the production object store setup on the dev with the aim of migrating this to a distributed object store with new datasets being stored by UUID. The old directories would have weight 0 meaning that nothing would be written to them.

Slugger70 (approved):

PR #888, opened 2022-08-01
Uom vet alphafold tpv rules

cat-bro:
Changes for new alphafold pulsar and group 'UoM_Vet_Alphafold'

I believe that if a user is a member of both 'Alphafold' and 'UoM_Vet_Alphafold' then the last matching rule will apply and their jobs will go to the 'UoM_Vet_Alphafold' pulsar.

The toolbox filter will require a playbook run in order to work for members of 'UoM_Vet_Alphafold' that are not already members of 'Alphafold'.


cat-bro:
@Slugger70 I've assumed that every slurm has a 'main' partition that is the default, does this sound right?

Slugger70 (approved):

PR #892, opened 2022-08-01
Fix alphafold rule for members of both groups

cat-bro:
As a member of both groups I cannot run alphafold


cat-bro:
There is probably a nicer way to write this, Bjoern said that roles are included in the job context so I'll play with this on dev ... in the meantime I'll merge this because I don't want to leave it broken.

PR #906, opened 2022-08-04
Set tmp_dir: True for all destinations, meaning temp files will be in subdirectories of the job working directories

cat-bro:
At the moment this is used for `__DATA_FETCH__` which means that tool no longer causes the worker tmp disks to overflow.

I can't think of what could go wrong with this, however

@Slugger70 would there be a greater overhead for the workers storing temp things on the NFS mounts instead of on the mounted /mnt/tmpdisk disks?


Slugger70:
It will likely make some things slower and job nfs will be busier but that's haow it will have to be

Slugger70 (approved):

PR #910, opened 2022-08-05
try removing volume specification for cactus tools

TomHarrop:
@cat-bro trying cactus without `$defaults`


cat-bro (approved):

cat-bro:
mem is inherited from the default tool if it's not specified, so it is cores * multiplier

PR #932, opened 2022-08-18
Common ansible env for jenkins jobs, add config_backup playbook

cat-bro:
Add config backup playbook to this repo. It has been running in its own gitlab repository but the hosts file goes out of date every time an IP address changes so it's better to use the source-of-truth host file in this repo.  The backed up files still go to the 'configs' gitlab repository.

Every jenkins ansible job can use the same virtual environment (/var/lib/jenkins/jobs_common/.ansible_venv), previously each one had its own venv.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 .vault_password
 venv
 ansible.cfg
-ansible_config_backup/*
 terraform/terraform*
 terraform/.terraform
 terraform/prod/.terraform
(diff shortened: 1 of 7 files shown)


Slugger70:
Where are these backups going to be stored? They will need to be private due to connection strings etc.

cat-bro:
same gitlab repo where they are currently stored.

Slugger70 (approved):

PR #949, opened 2022-09-01
add new dev worker and slurm partitions s3fs and cvmfs

cat-bro:
Add dev-w2 and give slurm 2 additional partitions. Partitions now are
```
main: dev-w1
cvmfs: dev-w1
s3fs: dev-w2
```

### DIFF
--- a/dev-workers_playbook.yml
+++ b/dev-workers_playbook.yml
@@ -11,15 +11,23 @@
       - name: Attach volume to instance
         include_role:
           name: attached-volumes
+      - name: Run the common role first
+        include_role:
+          name: common
+      - name: Include the cvmfs role
+        include_role:
+          name: galaxyproject.cvmfs
+        when: "{{ hostname.startswith('dev-w1') }}"
+      - name: Include the cvmfs role
+        include_role:
+          name: galaxyproject.s3fs
+        when: "{{ hostname.startswith('dev-w2') }}"
   roles:
-      - common
       - insspb.hostname
       - geerlingguy.pip
       - mounts
       - galaxyproject.repos
       - galaxyproject.slurm
-      - galaxyproject.s3fs
-      - galaxyproject.cvmfs
       - geerlingguy.docker
       - gantsign.golang
       - cyverse-ansible.singularity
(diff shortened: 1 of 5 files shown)


dev-workers_playbook.yml
```diff
@@ -11,15 +11,23 @@
...
         include_role:
           name: attached-volumes
+      - name: Run the common role first
+        include_role:
+          name: common
+      - name: Include the cvmfs role
+        include_role:
+          name: galaxyproject.cvmfs
+        when: "{{ hostname.startswith('dev-w1') }}"
+      - name: Include the cvmfs role
```
nuwang, line 21:
```suggestion
      - name: Include the s3fs role
```

group_vars/dev_workers.yml
```diff
@@ -22,3 +22,22 @@ docker_users:
...
 docker_daemon_options:
   data-root: /mnt/docker-data
+
+# cvmfs
+cvmfs_cache_base: /mnt/var/lib/cvmfs
+cvmfs_quota_limit: 20000
+
+# s3fs
+s3fs_cache_base: /mnt/var/lib/s3fs
+s3fs_ensure_diskfree: 80000
```
nuwang, line 32: What else would be writing to this disk? If temp files are going here, then the s3fs cache size would be substantially lower than cvmfs?

nuwang (commented):
LGTM! Some minor comments.

dev-workers_playbook.yml
cat-bro: :)

group_vars/dev_workers.yml
cat-bro: At the moment there is 95083MB free on the disk.  The majority of tmp files will be contained in /mnt/galaxy/tmp/job_working_directory (NFS) instead of on this disk.  Assuming no docker/singularity enabled tools used in testing, nothing else should be writing to this disk.  So maybe it should be `s3fs_ensure_diskfree: 75000`.

Slugger70 (approved):

PR #952, opened 2022-09-01
Reduce resources for some high-mem tools

cat-bro:
Reduce the cores/mem allocations for some high-mem tools to reduce congestion in the high-mem queues.  @Slugger70, @jlqfab what do you think?  We have sacct data but I'm not sure how to tell how efficient the allocations have been for the different tools.

```
halved:
flye: 120 -> 60
hifiasm: 120 -> 60
racon: 120 -> 60
medaka_consensus_pipeline: 120 -> 60
seqtk_seq: 120 -> 60
raven: 120 -> 60
hisat2: 60 -> 30
busco: 60 -> 30
spades tools: 60 -> 30

quartered:
minimap2: 120 -> 30

taken off high-mem altogether:
quast: 60 -> 16
```

### DIFF
--- a/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
+++ b/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
@@ -693,8 +693,8 @@ tools:
     - if: 1 <= input_size < 10
       cores: 16
     - if: input_size >= 10
-      cores: 120
-      mem: 1922
+      cores: 60
+      mem: 961
       scheduling:
         accept:
         - high-mem
@@ -707,8 +707,8 @@ tools:
     - if: 0.05 <= input_size < 5
       cores: 8
     - if: input_size >= 5
-      cores: 120
-      mem: 1922
+      cores: 60
+      mem: 961
       scheduling:
         accept:
         - high-mem
@@ -888,8 +888,8 @@ tools:
     - if: 0.5 <= input_size < 15
       cores: 8
     - if: input_size >= 15
-      cores: 60  # one quarter of a QLD high mem pulsar
-      mem: 961
+      cores: 30
+      mem: 480.5
       scheduling:
         accept:
         - high-mem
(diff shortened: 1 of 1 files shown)


jlqfab:
HI Catherine. I've looked at memory use data in sacct for the first 4 of these. There was no MaxRSS (max memory used) for any of these except hifiasm which showed (on pulsar-qld-high-mem0) usage of ~107.
We could drop memory assigned for all of some of these, and see how they go - sacct has a '--state OUT_OF_MEMORY' query that we could use to pick up which jobs are failing due to being out of memory.

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
```diff
@@ -693,8 +693,8 @@ tools:
     - if: 1 <= input_size < 10
       cores: 16
     - if: input_size >= 10
-      cores: 120
-      mem: 1922
+      cores: 60
```
Slugger70, line 696: Leave this one at 120 and 1922.

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
```diff
@@ -1020,8 +1020,8 @@ tools:
     - if: 5 <= input_size < 20
       cores: 16
     - if: input_size >= 20
-      cores: 120
-      mem: 1922
+      cores: 30
```
Slugger70, line 1023: This is another heavily cpu bound tool (actually all of them are..) so we should do some investigation into their scaling (on a tool by tool basis) before we make these changes. Are they linear? Are they exponential? What is the point of diminishing returns?

Slugger70 (changes requested):

Slugger70:
We should look at all of the tools we send to high mem and investigate their scaling with resources and make sensible decisions based upon that rather than across the board reduction.

PR #953, opened 2022-09-02
Stop using AWS as rw dir in object store, omit /mnt/test_mount from singularity volumes

cat-bro:
On dev, the aws storage is causing pulsar jobs with large outputs to fail.  Including the path to it in singularity_volumes causes all singularity jobs on slurm to fail.

There may be workarounds, this PR is just to stop using it as the primary storage for now.  Any singularity jobs using files created in the past two weeks will not work unless the datasets are copied.

### DIFF
--- a/host_vars/dev.usegalaxy.org.au.yml
+++ b/host_vars/dev.usegalaxy.org.au.yml
@@ -250,7 +250,7 @@ slurm_singularity_volumes_list:
   - "/mnt/galaxy/data:ro"
   - "/mnt/galaxy/data-2:ro"
   - "/mnt/galaxy/data-3:ro"
-  - "/mnt/test_mount:ro"
+  # - "/mnt/test_mount:ro"  # including this causes singularity jobs to fail
   - "/cvmfs/data.galaxyproject.org:ro"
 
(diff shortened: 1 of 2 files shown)


nuwang:
@cat-bro What size would be a large output? Over 5GB? AWS has a part size of 5GB.

cat-bro:
24G.  There is no need to troubleshoot this scenario because if we ever used the aws dir for storage in production it would be read-only, pulsars would not be writing their data to it.

I wonder what would happen if the test_mount directory were mounted within a containing folder, then the path would be `/mnt/intermediate_dir/test_mount` and the singularity volume could be `/mnt/intermediate_dir`, just a normal folder.  Maybe that would work.

cat-bro:
This object store could be using `<backend type="s3">` instead of treating the aws bucket as disk, see sample file
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/object_store_conf.xml.sample#L135

cat-bro:
The singularity problem turned out to be because the disk was not mounted on the workers, resolved with #958

PR #954, opened 2022-09-02
Update gxadmin-local.sh

cat-bro:
Add a line so that the command `gxadmin local` will list local functions. Update the docs so that each function has a description when listed. Fix some typos.


jlqfab (approved):
Looks good!

PR #956, opened 2022-09-05
schedule bionano_scaffold to slurm only

cat-bro:
There are errors running this on pulsar, it is OK on slurm.


jlqfab (approved):
All good.

PR #962, opened 2022-09-07
Add two reservation destinations to dev, set them to offline

cat-bro:
Add two destinations to dev for pulsar reservations.  These have very generic entries in the job conf, finer settings for individual tools with be in TPV.

For the TPV destinations there is a separate file which might make it easier for an automatic process to be switching them on/off or inserting context variables for expiry date/time.

When a reserved pulsar is spun up it could be assigned to one of these pulsar destination names and given whatever combination of rabbit username/vhost/password goes with that pulsar.


Slugger70 (approved):
I like this idea

PR #966, opened 2022-09-13
configure apt to set erlang version to fixed - no updates

jlqfab:
Configure apt to set erlang version to fixed (hold) so that it doesn't get updated. As per request from @Slugger70

### DIFF
--- a/roles/jasonroyle.rabbitmq/tasks/Debian/install.yml
+++ b/roles/jasonroyle.rabbitmq/tasks/Debian/install.yml
@@ -11,6 +11,11 @@
   apt:
     name: erlang
 
+- name: hold erlang version
+  dpkg_selections:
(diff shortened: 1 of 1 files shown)


cat-bro:
What happens if someone using this role does want to update the `rabbitmq_version` variable to a version too new for the originally installed erlang? (https://www.rabbitmq.com/which-erlang.html)

jlqfab:
One way is to explicitly install a particular version of Erlang, as is done with rabbitmq_version. Alternatively, if the rabbitmq_version is changed we can unhold the version of erlang and install the latest:
```
- name: unhold erlang version
  dpkg_selections:
    name: erlang
    selection: install

- name: install latest version of erlang
  apt:
    name: erlang
    state: latest
```

PR #977, opened 2022-09-20
An attempt to fix db backup swift woes

Slugger70:


Slugger70:
This gets out of sync too easily. @jlqfab - I might get you to work on how we can do this better.

PR #983, opened 2022-09-21
Change = in galaxy.j2 template to ==

cat-bro:
I thought single equal was right for this but ansible templating failed.  It worked when changed to == and the resulting file on dev looks right


Slugger70:
That makes sense

Slugger70 (approved):

PR #1006, opened 2022-10-07
Add webdav upgrade to aarnet playbook, add terms_url to galaxyservers.yml

cat-bro:
Apply Nuwan's webdav fix (tested on dev) to staging and aarnet playbooks.

Add `terms_url` to galaxyservers.yml.  This means there will be an extra menu item 'Terms and Conditions' and will be the value used in validation emails if email validation is switched on.  Tested on dev.

### DIFF
--- a/aarnet_playbook.yml
+++ b/aarnet_playbook.yml
@@ -64,6 +64,11 @@
       package:
         name: slurm-drmaa1
         state: latest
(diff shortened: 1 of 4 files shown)


cat-bro:
ref https://github.com/usegalaxy-au/infrastructure/pull/978

Slugger70 (approved):

PR #1023, opened 2022-10-17
Dev handler split

Slugger70:
This PR splits the dev server into 2 VMs - one for web handlers and one for job handlers and celery. Celery is also enabled to carryout tasks and there is a monitoring server on it called "Flower"

Dev is already running this system and is operational. There is still one mistake - the default server location needs to be updated in the resulting gravity supervisor configs. I will be trying to figure this out tomorrow.


cat-bro (approved):
This look great and very interesting.

For future I'm wondering if it's possible to define the host_galaxy_config in one place - the gravity sections are not exactly the same but could be handled with variables.  It would be very easy for the galaxy.yml variables in host_vars/dev and host_vars/dev-handlers to go out of sync

PR #1028, opened 2022-10-23
Fix tiaas vars

neoformit:
Prod TIaaS is not loading static files because the `tiaas_info.domain` variable has been set incorrectly (not your fault really, the name is misleading - fixed in new release).

[Take a look](https://usegalaxy.org.au/tiaas/) and you'll see that static files are not loading. I'm not sure why this stopped working now but this PR should fix it.


cat-bro (approved):

neoformit:
@cat-bro I SSH'd in and fixed this manually so no rush to run the playbook to fix.

PR #1031, opened 2022-10-24
Error retry values for slurm job runner

cat-bro:
The number of retries for InvalidJobException and InternalException defaults to 0, which is why the jobs failed so quickly this morning with this error in the handler logs
```
galaxy.jobs.runners.drmaa WARNING 2022-10-23 21:39:04,495 [pN:handler_0,p:2426728,tN:SlurmRunner.monitor_thread] (4912303/674752) unable to check job status because of InternalException exception for 1 consecutive tries: code 1: slurm_load_jobs error: Resource temporarily unavailable,job_id: 674752
galaxy.jobs.runners.drmaa WARNING 2022-10-23 21:39:04,495 [pN:handler_0,p:2426728,tN:SlurmRunner.monitor_thread] (4912303/674752) job will now be finished OK
```

 I've copied params from Galaxy Main's job conf where they set the value to 5 for each kind of error.

### DIFF
--- a/templates/galaxy/config/aarnet_job_conf.yml.j2
+++ b/templates/galaxy/config/aarnet_job_conf.yml.j2
@@ -9,6 +9,8 @@ runners:
     workers: 4
   slurm:
     load: galaxy.jobs.runners.slurm:SlurmJobRunner
+    invalidjobexception_retries: 5
+    internalexception_retries: 5
   pulsar_rest:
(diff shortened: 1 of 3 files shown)


cat-bro:
We could set values for invalidjobexception_state and internalexception_state (both 'ok' by default) to 'error', but it seems more important to make sure they will try again.

Slugger70 (approved):

PR #1045, opened 2022-11-08
Add line to tiaas extra config

cat-bro:
from CH request in infrastructure channel


jlqfab (approved):
By default tiaas auto-redacts contact details of trainers after 90 days. This setting disables that. Change request originates with Gareth and Igor.

PR #1052, opened 2022-11-10
Remove line in fstab to make training VMs reboot safe

cat-bro:


jlqfab (changes requested):
this will fix the issue - but would be good add something about `x-systemd.requires=cloud-init.service` as this is the real issue, `/dev/vdb` just happens to be the device chosen
I’m just thinking that if we come across the issue elsewhere in future, there will be something we can refer to.

jlqfab (approved):

PR #1064, opened 2022-11-16
on qld-pulsar-himem-2 limit number of slurm cgroup processes to preve…

jlqfab:
…nt mothur forking to infinity

### DIFF
--- a/pulsar-qld-high-mem_playbook.yml
+++ b/pulsar-qld-high-mem_playbook.yml
@@ -62,3 +62,17 @@
       become: yes
       become_user: root
       when: links.stat.islnk is defined and not links.stat.islnk
+    - name: limit processes spawned by slurm jobs on qld-pulsar-himem-2
+      #  limit number of slurm cgroup processes to prevent mothur forking to infinity
+      lineinfile:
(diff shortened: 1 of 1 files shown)


Slugger70 (approved):

pulsar-qld-high-mem_playbook.yml
```diff
@@ -62,3 +62,17 @@
       become: yes
       become_user: root
       when: links.stat.islnk is defined and not links.stat.islnk
+    - name: limit processes spawned by slurm jobs on qld-pulsar-himem-2
+      #  limit number of slurm cgroup processes to prevent mothur forking to infinity
+      lineinfile:
+          path: /lib/systemd/system/slurmd.service
+          regexp: "^TasksMax="
+          line: "TasksMax=2000"
+      when: "{{ hostname.endswith('himem-2') }}"
```
cat-bro, line 71: if hostname is set in all.yml to always be something.usegalaxy.org.au, this when will never be true https://github.com/usegalaxy-au/infrastructure/blob/master/group_vars/all.yml#L2

PR #1082, opened 2022-11-23
Update stats role and pulsar-azure IP

cat-bro:
Update get_queue_size.py and secrets.yml templates
Update SINFO to allow 24 chars for hostnames

Update pulsar-azure in hosts and playbook to comment out everything but pulsar and slg.galaxy_stats roles


cat-bro:
The reason for the update in get_sinfo is that pulsar-azure has a bunch of workers with disparate names (azure-gpu1, 10.0.0.4, small-1 etc)

Slugger70 (approved):

PR #1088, opened 2022-12-01
Adjust flye and training rule

cat-bro:
Adjust tpv config so that 0.5-10G flye jobs have 16 cores and the pulsar-training-large jobs are given the minimum of 16 cores and the number of cores they would have outside of training groups.

The pulsar-training-large change from 8 to 16 cores is a temporary fix for today's workshop.  There are ways the roles.yml config could be changed to make it more flexible but I favoured this solution for now because it won't break anything.


cat-bro:
@TomHarrop would this suit you? It ought to mean flye jobs and probably meryl jobs would have 16 cores

TomHarrop:
Yes please, this looks good to me.

jlqfab (approved):

PR #1091, opened 2022-12-05
set pulsar-qld-high-mem2 offline

cat-bro:


cat-bro:
@nuwang I remember having a conversation with you about where we put `- offline` in these.  For GA's purposes it is appropriate to put offline in the positive tags (accept, prefer, require), so that there are still test users (such as `pqld@usegalaxy.edu.au`) that can still run jobs to test the destinations when they are not available to regular users.

nuwang (approved):

nuwang:
Makes sense. We can use it in two ways I suppose. Add it to reject when we want full rejection, including test jobs. Use cases might be when the node is completely offline. We can then add offline to required when the node is ready, but needs to be tested first.

PR #1095, opened 2022-12-07
Use fork of ansible-galaxy-subdomains until next release

cat-bro:
The current version of ansible-galaxy-subdomains has a bug that breaks galaxy.  Switch version in requirements.yml to a forked copy of the role with the bug fix until the next release of this role (0.0.13) or later.  I've made a PR to their repo and I'll switch this requirement back as soon as it is fixed. This only affects dev.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -64,4 +64,8 @@
 - name: usegalaxy_eu.flower
   version: 0.3.1-alpha
 - name: usegalaxy_eu.galaxy_subdomains
-  version: 0.0.11
+  # version: 0.0.12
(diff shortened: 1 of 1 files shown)


nuwang (approved):

hexylena:
Very happy to:

- grant you push access to that role (y'all're the biggest user for now, EU hasn't switched)
- which in turns grants you release access (there's a github action, you just trigger the action, or manually push/create a tag)

so you don't have to maintain a fork, if y'all like :)

nuwang (approved):

PR #1098, opened 2022-12-09
pulsar-paw online

cat-bro:


jlqfab (approved):
good for now, but is the mount point going to change again? :-\

cat-bro:
@jlqfab I think the discrepancy that existed between fstab and reality probably comes down to human error somehow. The volume and workers had to be rebuilt a few months ago.

PR #1099, opened 2022-12-10
define dev job destinations in tpv instead of job conf

cat-bro:


nuwang (commented):
This looks good!
I noticed that there were a few user related conditions like:
```
    rules:
      - if: user.email == 'star@email.com'
        cores: 8
      - if: user.email != 'star@email.com'
        cores: 8
        scheduling:
          require:
            - pulsar
            - pulsar-nci-test
```
Perhaps, we can now add them directly as a user rule instead of a conditional?

nuwang:
@cat-bro The abstract property will require a switch to this branch: https://github.com/galaxyproject/total-perspective-vortex/pull/64
but it includes all commits from this PR: https://github.com/galaxyproject/total-perspective-vortex/pull/63

files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/vortex_config.yml
```diff
@@ -420,89 +420,3 @@ users:
         params:
           nativeSpecification: "--nodes=1 --ntasks={cores} --ntasks-per-node={cores} --mem={round(mem*1024)} --partition=azuregpu1"
           submit_native_specification: "--nodes=1 --ntasks={cores} --ntasks-per-node={cores} --mem={round(mem*1024)} --partition=azuregpu1"
```
nuwang, line 422: @cat-bro Can we override only the partition context variable for this? I'm not sure if that will work, but it should?

cat-bro: yes we can. That looks like old-fashioned code

nuwang (approved):
LGTM!

PR #1110, opened 2022-12-15
Dedup dev server variables

cat-bro:
A lot of variables for dev are in two places (host_vars/dev.usegalaxy.org.au and host_vars/dev-handlers.usegalaxy.org.au) and it is difficult to keep these in sync.  This changes host_vars/dev-handlers to be a small file with only what needs to be added/overwritten from the contents of host_vars/dev, then host_vars/dev is included in vars files for the dev-handlers playbook.


jlqfab (dismissed):

nuwang (approved):
Needs a rebase I guess.

jlqfab (approved):

PR #1126, opened 2023-01-05
add memory

TomHarrop:
I think my test job ran out of RAM. The same job completes on my desktop which has 64 GB.


cat-bro:
The biggest pulsar worker on dev only has 48G so it might be easier to test the tool on production as Cam has done here https://github.com/usegalaxy-au/infrastructure/pull/1116/files.  The folder /mnt/galaxy/local_tools on aarnet can be written to by all users.  The "test" label means the tool is only visible to us.

TomHarrop:
Thanks, I've added the tool wrapper manually to aarnet

cat-bro (approved):

PR #1129, opened 2023-01-09
Remove unused playbooks, hosts and files for DR, pulsar-eu-gpu and test-db

cat-bro:
Remove some unused hosts and files.

For the DR setup, these playbooks were copied from production playbooks in late 2021 and are out of date. If a DR playbook were needed it would be better to start from scratch.


nuwang (approved):
LGTM!

PR #1135, opened 2023-01-12
rules for tool hifiasm_meta

TomHarrop:


cat-bro:
@TomHarrop would you be happy with `singularity_enabled: true` for this? It's using a single conda package there will (i think) be a container

TomHarrop:
Yep! I do my testing on biocontainers rather than conda, so this tool has already been tested with Singularity.

cat-bro (approved):

Issue #1117, opened 2022-12-20
Move singularity caches from root disks


cat-bro:
The singularity application's cache dir still defaults to /home/galaxy/.singularity on the galaxy head node.  This could probably be resolved by using the `SINGULARITY_CACHEDIR` environment variable and setting it to `/mnt/singularity_data` (or something) in the aarnet [host file](https://github.com/usegalaxy-au/infrastructure/blob/master/host_vars/aarnet.usegalaxy.org.au.yml#L153)

The slurm workers also have singularity caches on the root disk but this could be resolved within job conf/tpv.

PR #1136, opened 2023-01-13
galaxy singularity cache data to go to /mnt/singularity_data on dev

cat-bro:
Set value for SINGULARITY_CACHEDIR on dev so that space is not taken up in /home/galaxy/.singularity.

Every time a new tool is run with singularity it is causing a fair bit of space to be used on the head node root disk, so if this setup works I'll copy it on staging + production.

path name is '/mnt/singularity_data' so that it sits alongside '/mnt/docker_data' (the place that docker things go to instead of the root disk)

### DIFF
--- a/dev_playbook.yml
+++ b/dev_playbook.yml
@@ -25,6 +25,13 @@
         owner: root
         group: galaxy
         mode: 0775
+    - name: create dir for galaxy's singularity cache
+      file:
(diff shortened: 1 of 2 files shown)


cat-bro:
ref #1117

jlqfab (approved):
Looks good!

PR #1144, opened 2023-01-17
Param singularity on ebi-gxa scanpy tools

neoformit:
Not sure if you can do this? :shrug: 
We want to singularity-enable all `ebi-gxa/scanpy*` tools right?


cat-bro:
this is python regex and not a glob, so needs to be `toolshed.g2.bx.psu.edu/repos/ebi-gxa/scanpy.*` instead of `toolshed.g2.bx.psu.edu/repos/ebi-gxa/scanpy*`

Could you please add the same entry to staging/vortex_config.yml?  (This duplication will no longer be necessary once tpv is upgraded to version 2)

cat-bro (approved):
Thanks @neoformit

PR #1148, opened 2023-01-20
Update dev tpv to version 2.0.1

cat-bro:
@nuwang can you please check the change to destinations.yml.j2? By chaging max_accepted_cores/max_accepted_mem to max_cores/max_mem should jobs with larger cores/mem be scaled down to what is available?


nuwang:
Both are likely to be needed. `max_accepted_cores` is needed to determine whether the destination matches in the first place. If it is left undefined, it will always be a match of course. Once matched, the max_cores will clamp it down to the desired range. I assume Slurm shouldn't match everything, right?

cat-bro:
Because it's dev I'm happy for it to run every kind of job on the 16 core node. Same with staging. It's not something we'd have on production.

nuwang (approved):
LGTM!

PR #1155, opened 2023-01-25
Add hosts to hosts template file on aarnet VMs

cat-bro:
Put the 'add hosts' tasks in their own file and run them twice if a variable 'hosts_template_file' is defined.


jlqfab (approved):
nice!

PR #1157, opened 2023-01-25
upgrade staging tpv to 2.0.1

cat-bro:
This updates staging's config to use the production tools.yml file in addition to its own files. The advantage of doing so is that job running preferences (singularity, docker) will only need to be set in one place.


templates/galaxy/config/staging_job_conf.yml.j2
```diff
@@ -35,60 +35,10 @@ execution:
...
-      outputs_to_working_directory: false
-      docker_enabled: true
-      docker_volumes: $defaults
-      docker_sudo: false
-      docker_net: bridge
-      docker_auto_rm: true
-      docker_set_user: ''
-      require_container: true
-      container_monitor_result: callback
+        - "{{ galaxy_dynamic_job_rules_dir }}/total_perspective_vortex/staging_tools.yml"
```
nuwang, line 38: Shouldn't there be a reference to the shared database and production tools file?

cat-bro: Good point, I will add the shared database

staging_tools.yml is a symlink (in the infrastructure repo) to the production tool file, so copies to the staging VM with the same name.  The way that the galaxyproject.galaxy role handles job rules files makes it difficult to list a file in a different directory, hence the symlink.  staging_vortex_config.yml defines the default tool for staging with weighted random sampling as the rank function.

nuwang (approved):

Issue #1142, opened 2023-01-16
tpv ranking function discussion


cat-bro:
The ranking function defined in [default_tools.yml.j2](https://github.com/usegalaxy-au/infrastructure/blob/master/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/default_tool.yml.j2#L20) queries the influx API and receives info on the core utilisation (percentage)  for each destination under consideration.  If it is choosing between slurm and pulsar-mel2 it will choose the one with lowest proportion of cores currently being used.  This has worked pretty well but there are some issues:
- In the event that a job's datasets take a few minutes to get to a pulsar destination, the job will not start and the cores will not be used until the datasets arrive, so there is a lag between a job being assigned and the cores being utilised. This can lead to destinations being oversubscribed when others are available.
- The ranking is considering the relative amount of nodes available rather than the absolute number, which makes it more likely for a small destination like pulsar-mel2 (40 cores) to be oversubscribed.
- The API does not make memory data available.  As long as jobs always have 3.8 Gb memory per core this is OK, but we are moving away from this by using the shared database, and there is no longer any expectation of a fixed ratio between cores and mem. It will no longer be enough to be using core allocation info to choose between destinations.

Simon's intention was to try out using galaxy database data for ranking instead of data from the stats API.  Every job has cores and mem data available in the destination_params field. It would make the allocation data more accurate and enable us to take into account both cores and mem allocated to a destination. It could also put pressure on the database, though Simon's initial tests of the query suggested that it wouldn't.

So there are several possibilities for updating this including:
- tpv queries database during ranking and no API calls are needed
- stats API endpoint includes data from the galaxy db
- add using memory and queued jobs to the stats API info (from slurm) and use this in the ranking function

cat-bro:
Things that we can get from slurm data but not from galaxy db are
(1) The breakdown of space available on each worker node for a destination
(2) Whether all of the worker nodes are running

PR #1158, opened 2023-01-30
Production tpv upgrade

cat-bro:
This is a draft and is still missing a few things.  There are some TODO items in destinations.yml.j2 for things that need to be checked.

All tools will need to have mem values added, for using with shared db.

The rank function needs updating to use with shared db (ref #1142 )

The current function that collects tools by destination for grafana relies on destinations being defined in job_conf.yml and this will need to be updated before this PR can be merged.


cat-bro:
Tool specifications in tools.yml have previously had lowest cores/mem as default and higher resource allocations set in conditional rules. This changes to have the non-failing condition with the highest cores/mem as default and lower resource allocations set in conditional rules, e.g. deeptools_plot_heatmap

```
  toolshed.g2.bx.psu.edu/repos/bgruening/deeptools_plot_heatmap/deeptools_plot_heatmap/.*:
    cores: 3
    rules:
    - if: input_size >= 0.3
      cores: 7
```
changes to 
```
  toolshed.g2.bx.psu.edu/repos/bgruening/deeptools_plot_heatmap/deeptools_plot_heatmap/.*:
    cores: 7
    mem: 26.9
    rules:
    - id: deeptools_plot_heatmap_small_input_rule
      if: input_size < 0.3
      cores: 3
      mem: 11.5
```

Using this convention should make it easier to switch to using a shared database rule for a tool.  If we wanted to use shared_db for deeptools_plot_heatmap we could either delete the entry entirely or delete {cores: 7, mem: 26.9} from the above.

cat-bro:
This PR doesn't change the rank function in any way but I do think it needs to change.  I'm working on alternatives to the rank function on dev. We can swap the rank function later on when we have seen what the new edge cases will be on production.

.github/workflows/check_tpv_config.yml
```diff
@@ -31,26 +31,68 @@ jobs:
...
+          ansible_command = f'ansible-playbook -i {hosts_file} {playbook_file} --extra-vars \'{extra_vars}\''
+          try:
+            subprocess.check_output(ansible_command, shell=True)
+          except subprocess.CalledProcessError as e:
+            print(e.output)
+            raise Exception('Failed to template {os.path.basename(fname)}')
+          return output_fname
+
+        def sanitize_text(fname):
+          with open(fname) as handle:
```
nuwang, line 77:
```suggestion
        def sanitize_text(fname):
          """Replaces any references to vault variables with a constant string as the
               vault is not decrypted during tests"""
          with open(fname) as handle:
```

.github/workflows/check_tpv_config.yml
```diff
@@ -31,26 +31,68 @@ jobs:
...
-            except KeyError as e:
-                raise Exception(f'Destination id {vd} not found in job config file')        
       shell: python
-    - name: Run tpv lint on all vortex config files
-      env:
-        SETTING: github
-      run: for f in files/galaxy/dynamic_job_rules/production/total_perspective_vortex/*; do tpv lint $f; done
\ No newline at end of file
+      run: |
+        import re
```
nuwang, line 42:
```suggestion
        """This code uses ansible to template out the jinja templates for tpv configs and job config,
              then runs the tpv linter on the resultant files."""
        import re
```

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/roles.yml
```diff
@@ -1,41 +1,62 @@
...
+        exclude_tags = ['training-exempt']
+
+        def tag_values_match(entity, match_tag_values=[], exclude_tag_values=[]):
+          return (
+              all([any(entity.tpv_tags.filter(tag_value=tag_value)) for tag_value in match_tag_values])
+              and not any([any(entity.tpv_tags.filter(tag_value=tag_value)) for tag_value in exclude_tag_values])
+          )
+
+        tag_values_match(entity, include_tags, exclude_tags)
+        # helpers.tag_values_match(entity, include_tags, exclude_tags)
```
nuwang, line 56: We can make a release of tpv immediately so you can use this instead?

.github/workflows/check_tpv_config.yml
cat-bro: Thank you!

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/roles.yml
cat-bro: That would be good!

nuwang (dismissed):
LGTM!

PR #1165, opened 2023-02-06
Remove max_quant_w8 partition and limit maxquant in tpv

cat-bro:
The slurm partition max_quant_w8 started off as a priority queue for maxquant because no other jobs could run there. When galaxy became more busy, the w8 worker was added to the main queue so it became difficult for maxquant jobs to get a place to run. This PR removes the max_quant_w8 partition and puts some limits on the maxquant tool in tpv to stop it from taking over slurm.

Note: this has no effect the local maxquant_test tool which is subject to different rules.

### DIFF
--- a/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
+++ b/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
@@ -270,16 +270,18 @@ tools:
     cores: 6
   toolshed.g2.bx.psu.edu/repos/galaxyp/maxquant/maxquant/.*:
     cores: 16
-    context:
-      partition: max_quant_w8
+    rules:
+    - if: |
+        user_limit = 2
+        total_limit = 4
+        helpers.concurrent_job_count_for_tool(app, tool, user) >= user_limit or helpers.concurrent_job_count_for_tool(app, tool) >= total_limit
+      execute: |
+        from galaxy.jobs.mapper import JobNotReadyException
+        raise JobNotReadyException()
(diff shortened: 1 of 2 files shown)


files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
```diff
@@ -270,16 +270,18 @@ tools:
     cores: 6
   toolshed.g2.bx.psu.edu/repos/galaxyp/maxquant/maxquant/.*:
     cores: 16
-    context:
-      partition: max_quant_w8
+    rules:
+    - if: |
+        user_limit = 2
+        total_limit = 4
+        helpers.concurrent_job_count_for_tool(app, tool, user) >= user_limit or helpers.concurrent_job_count_for_tool(app, tool) >= total_limit
```
nuwang, line 277: This conditional seems to be common enough that perhaps we should move it into tpv core with something like:
```suggestion
        helpers.do_concurrent_jobs_exceed_limits(app, tool, user, user_limit=2, global_limit=4)
```
That function could handle return true or false, depending on whether only user_limit, global_limit or both are defined.

cat-bro: That's a good idea and saves it from doing the job query twice

cat-bro: I'd like to merge this as is since we are still on v1.x

nuwang (approved):

PR #1168, opened 2023-02-08
Add cron job for deleting old jwds from pulsar-azure

cat-bro:
ubuntu_maintenance_key public key has been added to pulsar-azure and this key will only be accepted for aarnet's IP.


cat-bro:
This is adding the same thing as already exists for the other pulsars. Every night it will check what's in the working directory on the pulsar against job states in the galaxy db and send a script to remove job working directories that are no longer needed.

nuwang (approved):

PR #1182, opened 2023-02-17
update max_map_count for maxquant on pqhm1

cat-bro:
Since pqhm0 is out of order, switch maxquant_test machine to pqhm1.


jlqfab (approved):
looks good

PR #1186, opened 2023-02-19
Update stats role

cat-bro:
Update stats role so that it can be included in playbooks. Also fix db connection string in another role that was requesting data from aarnet-db's public IP.


cat-bro:
The stats role has been there for a long time but was never finished. All of the stats collection scripts on aarnet and pulsars have been copied from other VMs.  These changes update the templates from values in the in-use scripts.

jlqfab (approved):

PR #1192, opened 2023-02-28
Add user-data-7 to aarnet-user-nfs

cat-bro:
Add /mnt/user-data-7 to aarnet-user-nfs and its nfs exports, mount it on aarnet head and workers

When user-data-6 is 90-95% full, user-data-7 should be added as the first entry in the object store config


nuwang (approved):
One thing we may need to do soon is to consolidate these object stores somehow. As the number of object stores increases, the performance decreases as I recall.

PR #1193, opened 2023-03-01
Create Makefile

neoformit:
@cat-bro I'm sure you have other things to add to this

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,18 @@
+# Management commands
+
+KEY_FILE = ~/.ssh/galaxy  # symlink this if your filename is different
+
+install-roles:
+	ansible-galaxy install -p roles -r requirements.yml
+
+update-roles:
+	python scripts/update_roles.py
+
+run-dev:
+	ansible-playbook -i hosts --key-file $(KEY_FILE) dev_playbook.yml
+
+run-staging:
+	ansible-playbook -i hosts --key-file $(KEY_FILE) staging_playbook.yml
+
+run-prod:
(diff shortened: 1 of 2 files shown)


cat-bro:
@neoformit this is a really good idea. It will simplify running playbooks and will make playbook automation easier.

Since there are maybe 30+ playbooks I'm wondering if the run command can be generalised, i.e. 'run-XYZ' or maybe 'run XYZ' would result in the running of 'XYZ_playbook.yml'.

neoformit:
Good idea - I'm not too familiar with Makefiles but I'll see if there's a way to parameterize it.

neoformit:
Standard use of `make` allows passing of named params, which might be too wordy for us:

```
make run book=dev
```

The following is more concise but is a hack and non-standard use of `make`...

```make
# If the first argument is "run"...
ifeq (run,$(firstword $(MAKECMDGOALS)))
  # get the second argument for "run"
  PLAYBOOK := $(wordlist 2, 2, $(MAKECMDGOALS))
  # ...and turn it into do-nothing target
  $(eval $(PLAYBOOK):;@:)
endif

run:
    ansible-playbook -i hosts --key-file $(KEY_FILE) $(PLAYBOOK)_playbook.yml
```

This allows running with a positional arg like:

```
make run dev
```

But it wouldn't handle mis-spelt playbook names elegantly. We'd be better off using a bash script for that.

cat-bro:
I love the non-standard use of Make option. I just copied it into my infrastructure folder and ran `make run dev-pulsar`. It is very cool.

neoformit:
I'll add that in :grin:

cat-bro (approved):

PR #1196, opened 2023-03-02
fix issue of destination of symlinks in home directories being chowned

jlqfab:
Rather than removing the code block that chown's the contents of user directories, simply don't follow symlinks. This prevents the destination of the symlink from being chown-ed.


cat-bro:
I'd prefer removing this task. The only reason for it was the user id change. There is no use case for it now. Backing up passwd isn't needed now either.

cat-bro (approved):

PR #1202, opened 2023-03-13
Put dev destination tags in host variables to be used by tpv and job conf

cat-bro:
This is a model for production destinations which have similar tags.


nuwang (approved):
Nice solution, because this makes it easier to keep the tags in sync. Out of curiosity, does the other solution of duplicating the destination names in job_conf also work? (While acknowledging the difficulty of keeping that in sync)

cat-bro:
Yes, duplicating the destination names in job conf does work as long as the tags are also duplicated. This PR is doing that by for-looping through job_conf_limits_environments in the job conf:
```
{% for env in job_conf_limits.environments|d({}) %}
    {{ env }}:
      tags: {{ job_conf_limits.environments[env].tags }}
{% endfor %}
```

PR #1204, opened 2023-03-16
Set galaxy log level to DEBUG

cat-bro:
The log level is currently at the highest level, TRACE. DEBUG is the default level. Options are (WARN, INFO, DEBUG, TRACE)

### DIFF
--- a/group_vars/galaxyservers.yml
+++ b/group_vars/galaxyservers.yml
@@ -131,7 +131,7 @@ galaxy_conda_exec: mamba
 group_galaxy_config:
   galaxy:
     check_migrate_tools: false
-    log_level: TRACE
+    log_level: DEBUG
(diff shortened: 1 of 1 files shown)


cat-bro:
I don't know if this will make a difference. I can't see TRACE level messages in the logs at all. Lots of DEBUG and INFO

nuwang:
Yes, I don't think there are many TRACE level messages. Do you want to try temporarily setting it to INFO and measuring impact? Also, checking whether logs at INFO level are good enough is also a useful data point.

nuwang (commented):
Yes, I don't think there are many TRACE level messages. Do you want to try temporarily setting it to INFO and measuring impact? Also, checking whether logs at INFO level are good enough is also a useful data point.

cat-bro:
This doesn't seem important now, am closing it.

PR #1206, opened 2023-03-17
Make quast a slurm-only tool

cat-bro:
Pulsar is not transferring all of the files back for quast jobs. This needs further investigation but in the meantime it should not run on pulsar.


cat-bro:
It never needed to be going to high-mem either... we put it there early on at somebody's request but 16 cores + 62G will be easily enough for the large jobs

nuwang (approved):

PR #1214, opened 2023-03-27
add singularity_enabled for kraken2tax

cat-bro:


cat-bro:
@jlqfab Galaxy has been trying to install the conda environment in the job working directory for each kraken2tax job. It fails, and tries several permutations before settling on an environment __gawk@_uv_ (containing the latest gawk). There is a GA user who runs many of these jobs at a time and it has an impact on the web server load (at a quiet time up to 7 from 1). I'll make a github issue to galaxy about this but we can avoid this behaviour by using singularity.

jlqfab (approved):

PR #1218, opened 2023-03-28
set pawsey file mounts to available after pawsey VM migration

jlqfab:


jlqfab:
pawsey-user-nfs VM has been migrated. I’ve remounted user-data-2,3 on aarnet head and workers.

cat-bro (approved):

PR #1224, opened 2023-03-31
 TPV entry for AlphaFold v2.3.1_2+

neoformit:
Drop AlphaFold version from `v2.3.1_2+` as will be set in tool XML from hereon.


cat-bro:
It's missing a `rules:` before the `- if`

neoformit:
How's that?

cat-bro (approved):

PR #1238, opened 2023-04-25
Add playbook for setting up tpv dry run from local files

cat-bro:
I'm adding this as a draft because there are already too many playbooks in the root directory. I might find another place for this.

This is a playbook that templates and copies all of the files needed for tpv dry run then creates a script run_dry_run.sh with job conf and tpv files as args to the tpv dry-run command, the user just needs to add the other arguments, e.g.
`bash run_dry_run.sh --tool=toolshed.g2.bx.psu.edu/repos/bgruening/hifiasm/hifiasm/0.19.3+galaxy0 --input-size 100`

Sharing with @kysrpex, @sanjaysrikakulam, @mira-miracoli (note that my vars will not work for EU, but something similar might)

### DIFF
--- a/tpv_dry_run_playbook.yml
+++ b/tpv_dry_run_playbook.yml
@@ -1,0 +1,57 @@
+- hosts: localhost
+  vars_files:
+    - group_vars/aarnet.yml
+    - group_vars/galaxyservers.yml
+    - host_vars/aarnet.usegalaxy.org.au.yml
+    - group_vars/VAULT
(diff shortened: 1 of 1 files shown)


sanjaysrikakulam:
Thank you @cat-bro!

mira-miracoli:
I think it would make sense to include that in the https://github.com/usegalaxy-eu/ansible-tpv-lint role,
since the script does basically the same but with linting (as far as I understand)

PR #1241, opened 2023-04-26
Add Helena's script for finding updated versions of ansible roles

cat-bro:
Add script from this gist: https://gist.github.com/hexylena/c60fe949dd76ce8e3116d2c745bfeb63

It goes through requirements.yml and for each role reports if there is a new version available.


nuwang (approved):
Looks good!

PR #1244, opened 2023-04-27
override shared db rule for sailfish

cat-bro:


jlqfab (approved):

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
```diff
@@ -201,6 +201,17 @@ tools:
   toolshed.g2.bx.psu.edu/repos/bgruening/salmon/salmon/.*:
     cores: 5
     mem: 19.1
+  toolshed.g2.bx.psu.edu/repos/bgruening/sailfish/sailfish/.*:
+    cores: 4
+    mem: 15.3
+    rules:
+    - id: sailfish_small_input_rule
```
nuwang, line 208: @cat-bro 4 and 15.3 is a pretty big difference from the shared db recommendation of 6 and 70. Is the shared db allocating too much?
https://github.com/galaxyproject/tpv-shared-database/blob/256a9659fd36b17fd919e4693097af9cf2ca50b8/tools.yml#L246

Also, should we consider moving the small input rule to the shared db?

PR #1251, opened 2023-05-04
Increase max_connections for staging db

cat-bro:
There have been problems with jobs on staging since upgrading to release 23.0. It is difficult to debug because the database  keeps running out of available connection slots.


nuwang (approved):
Interesting... 50 already seems like a decent number - given that we don't have that many processes/threads accessing the database?  An increase of 10 doesn't seem like much. We can try it I guess, but should it be bumped up to say 100? Prod appears to be using 200.

cat-bro:
50 has been enough for dev and staging up until now, and dev still seems OK. I'm not really sure how many connections we expect from the galaxy application. It may have increased with the new version. If this doesn't fix it I'll try 100.

PR #1254, opened 2023-05-05
Remove webdav patch from galaxy playbooks

cat-bro:


nuwang (commented):
I just realised that even though cloudstor is deprecated, we still have the griffith owncloud store: https://github.com/usegalaxy-au/infrastructure/blob/7597326b658d4389f3a7ab2ebe089a38449aba53/files/galaxy/config/file_sources_conf.yml#L12

I guess we may need to follow up on whether that's being used.

The dependency is actually coming through requests:
https://github.com/psf/requests/blob/2ad18e0e10e7d7ecd5384c378f25ec8821a10a29/setup.py#L64

If we simply remove `state: forcereinstall`, I think that should fix the problem for now.

cat-bro:
Why did it ever have 'state: forcereinstall' if it is ok without it? My purpose in making this PR is to stop the galaxy handlers from breaking every time the playbook is run. Please do update this if you have a solution, otherwise we should merge as is and update it later.

cat-bro:
Would it be better as a pre task so that galaxy's pinned requirement versions are installed later?

nuwang:
The `state: forcereinstall` is because Galaxy already installs it, and therefore, a pip upgrade is required to install the new version. Unfortunately, ansible's implementation appears to be upgrading dependencies too, although I don't think pip does that: https://stackoverflow.com/questions/2861183/upgrade-package-without-upgrading-dependencies-using-pip

This should be a better long term fix: https://github.com/usegalaxy-au/infrastructure/pull/1255/files

We could add it to a prehook, but that wouldn't work with a fresh install, since ansible-galaxy creates the virtualenv.

cat-bro:
> Unfortunately, ansible's implementation appears to be upgrading dependencies too, although I don't think pip does that

When I was installing the packages manually to see what was leading to the upgraded urllib, I installed this with `pip install --force-reinstall webdavclient3@git+https://github.com/ezhov-evgeny/webdav-client-python-3@0f17fa7946e66f7963db367d0d6b2e7f940ebeb8` assuming that it would do the same thing as the ansible, and it did upgrade urllib3 to 2.0.2.

> We could add it to a prehook, but that wouldn't work with a fresh install, since ansible-galaxy creates the virtualenv.

We can initialize the galaxy_virtual_env at any time and install packages before galaxy installs packages. We do this with the install-tpv role.

> This should be a better long term fix: https://github.com/usegalaxy-au/infrastructure/pull/1255/files

Thanks @nuwang, I'm closing this in favour of that one.

nuwang:
In that case, using the prehook may be even better.

I guess what we really want is the pip equivalent of upgrade, but I think ansible only provides force-reinstall?

PR #1264, opened 2023-05-15
Enable ITs

cat-bro:
Enable members of trusted_for_ITs and any training group to see/use these tools.

There is also the option of using only the first commit from this and making the interactive tools open to everyone.


nuwang (approved):

cat-bro:
This has been vetoed.

PR #1269, opened 2023-05-18
disable celery tasks on dev

cat-bro:
This was working on dev for 6 months or so: some small jobs such as uploads would be handled by celery instead of passed to slurm workers. Switching it off because jobs upload jobs are remaining in new state indefinitely. This is not used on staging/production.


nuwang (approved):
Has this been reported upstream? Marius is probably the person to inform.

PR #1271, opened 2023-05-19
Use process_manager: systemd for staging

cat-bro:
Use process_manager: systemd for staging: no longer using supervisor.

tusd is listed in the gravity section instead of it having a service set up by the ansible role. Gravity sets up the service galaxy-tusd and the old service tusd-main has been stopped and should be removed.

The path for the toolbox_filters folder in the galaxy codebase has changed mid-release.


nuwang (approved):
LGTM

PR #1272, opened 2023-05-22
log slow queries on aarnet db

cat-bro:


cat-bro:
This has been put in place on aarnet-db but never merged in

jlqfab (approved):

PR #1289, opened 2023-06-01
Send kraken2 and antismash to high memory pulsars, reduce some high memory allocations

cat-bro:
Send all kraken2 jobs to high memory pulsars because of a large db in cvmfs that is either new or people have only now started using it.

Antismash can run on pulsar and would also go to high memory pulsars.

Reduce allocations for some tools. Tools targeted for this are ones that have had 120 cpu / 2TB allocations. This is not across the board, some have been left as they are. It's necessary to reduce some of these because the high memory pulsars are prone to bottlenecks and a greater number of tools are going to need to run there.


cat-bro:
For the most part I am not reducing the allocations to what is in the shared DB. In some cases we have moved these tools to the high memory allocation at the request of people in the GA team. These can remain special without using 120 cpus per job which is unnecessary in most cases.

nuwang (dismissed):

PR #1303, opened 2023-06-06
galaxy etca backup and db playbooks

cat-bro:
Add playbooks and vars files for backup and db servers


cat-bro:
The galaxy db backup role and variables have been commented out so that they cannot be run until galaxy Sydney is production galaxy. When updating the values I noticed that one of the variables 'db_server' that I updated had been set to '192.168.199.56' which is not the internal IP for aarnet-db. The backup script on the aarnet-backup machine has the correct IP so these are out of sync. There could be other differences between what the role produces and the actual script used for backing up the database.

cat-bro:
the postgresql_object_users blocks in dev-db, staging-db and aarnet-db were so similar that they have been moved into dbservers.yml

cat-bro:
these commits are part of #1313: I am rolling this PR into that one

PR #1310, opened 2023-06-09
Webhooks role

neoformit:
- Create `webhooks` role that allows custom webhooks to be installed to galaxy servers
- Add `subdomain_switcher` webhook that allows users to switch between configured subdomains
- Add `webhooks` role and `subdomain_switcher` webhook to dev
- Role not yet tested on dev


roles/webhooks/main.yml
```diff
@@ -0,0 +1,11 @@
+# Copy local webhooks into webhooks folder
+- name: copy webhook plugins
+  copy:
+    src: "{{ webhook_plugins_src_dir }}/{{ item }}"
+    dest: "{{ webhook_plugins_dest_dir }}/{{ item }}"
+    owner: root
+    group: root
+    mode: 0755
+  loop: "{{ webhook_plugins }}"
```
cat-bro, line 9: It would make more sense for webhook_plugins to be `[]` instead of `null`, could leave out the `when` in this block.

roles/webhooks/main.yml
```diff
@@ -0,0 +1,11 @@
+# Copy local webhooks into webhooks folder
+- name: copy webhook plugins
+  copy:
+    src: "{{ webhook_plugins_src_dir }}/{{ item }}"
```
cat-bro, line 4: src: "{{ webhook_plugins_src_dir }}/{{ item }}.yml.j2"

roles/webhooks/main.yml
```diff
@@ -0,0 +1,11 @@
+# Copy local webhooks into webhooks folder
+- name: copy webhook plugins
+  copy:
```
cat-bro, line 3: template:

templates/galaxy/webhooks/subdomain_switcher/config.yml
```diff
@@ -0,0 +1,67 @@
+id: subdomain_switcher
```
cat-bro, line 1: by convention this filename would be 'config.yml.j2' because it is a template

roles/webhooks/main.yml
cat-bro: This probably wouldn't work because "{{ webhook_plugins_src_dir }}/{{ item }}" is a directory..

neoformit:
@cat-bro have refactored this to template all files for each `webhook_plugins` directory. Also removes webhook dirs from host that are not in `webhook_plugins`. Tested locally.

cat-bro (approved):
Looks good to me!

PR #1315, opened 2023-06-13
Update handlers in dev/staging update playbooks

cat-bro:


staging_update_configs_playbook.yml
```diff
@@ -9,13 +9,9 @@
...
       - secret_group_vars/ubuntu_maintenance_key
   handlers:
-    - name: restart handlers
-      become: True
-      become_user: galaxy
-      command: "{{ galaxy_venv_dir }}/bin/galaxyctl restart handler_0 handler_1 handler_2"
-      environment:
-        GRAVITY_STATE_DIR: "{{ galaxy_gravity_state_dir }}"
-        GALAXY_CONFIG_FILE: "{{ galaxy_config_file }}"
+    - name: galaxy gravity restart handlers
```
nuwang, line 12:
```suggestion
    - name: restart handlers
```

staging_update_configs_playbook.yml
```diff
@@ -9,13 +9,9 @@
...
+      command: "{{ galaxy_venv_dir }}/bin/galaxyctl -c {{ galaxy_config_file }} restart handler"
+      listen: "restart handlers"
```
nuwang, line 14: I don't see this listen group being used elsewhere, so should it just be using the name?

dev_update_configs_playbook.yml
```diff
@@ -12,25 +12,21 @@
...
   handlers:
     - name: galaxy gravity restart
-      become: True
-      become_user: galaxy
-      command: "{{ galaxy_venv_dir }}/bin/galaxyctl graceful"
-      environment:
-        GRAVITY_STATE_DIR: "{{ galaxy_gravity_state_dir }}"
-        GALAXY_CONFIG_FILE: "{{ galaxy_config_file }}" 
+      command: "{{ galaxy_venv_dir }}/bin/galaxyctl -c {{ galaxy_config_file }} graceful"
+      listen: "restart galaxy"
```
nuwang, line 16: The listen task is named differently elsewhere.
https://github.com/usegalaxy-au/infrastructure/blob/d193aadea3cfb7fa8acdfe354f4dac6d3f173230/group_vars/galaxyservers.yml#L25

nuwang (commented):
It looks like the handler tasks are redefined several times. Perhaps they can be centralized in a single handler include?

staging_update_configs_playbook.yml
cat-bro: I'm using the defaults from the galaxyproject.galaxy role https://github.com/galaxyproject/ansible-galaxy/blob/main/handlers/gravity_23.0.yml

I'm excluding become and become_user here because these playbook always run as root

dev_update_configs_playbook.yml
cat-bro: That variable is deprecated. https://github.com/galaxyproject/ansible-galaxy#optional-variables I'll remove it.

staging_update_configs_playbook.yml
cat-bro: I don't know if we can have a central handler for multiple playbooks, worth checking out but not today. For the production playbooks we can now use the defaults from the galaxyproject.galaxy role. These update playbooks do not use the galaxyproject.galaxy role.  There are three of them: one for dev/staging/production.

dev_update_configs_playbook.yml
nuwang: I also saw several other places that have "Restart Galaxy" instead of "restart galaxy"

staging_update_configs_playbook.yml
nuwang: I think that's an implementation detail and should ideally not be relevant to the core task of restarting handlers, so just thought I'd flag it. Fine with leaving it as is also.

dev_update_configs_playbook.yml
cat-bro: I can't find a single one with a text search

nuwang (approved):

PR #1319, opened 2023-06-13
specify grafana version for stats.usegalaxy.org.au so that it doesn't…

jlqfab:
… get updated when playbook run


cat-bro:
Thanks @jlqfab !

cat-bro (approved):

PR #1322, opened 2023-06-15
Deploy genomes lab to aarnet

neoformit:
FYI this will be the first time that a subdomain has been deployed to prod with this role.

### DIFF
--- a/aarnet_playbook.yml
+++ b/aarnet_playbook.yml
@@ -64,6 +64,7 @@
     - slg.galaxy_stats
     - galaxy-pg-cleanup
     - galaxyproject.tiaas2
+    - webhooks
   post_tasks:
     - name: Make local_tool directory group-writable by machine users
       file:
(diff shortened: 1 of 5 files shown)


cat-bro:
It looks like the subdomains role is missing (unless that has been absorbed into one of the other roles) as well as nginx changes, things for the galaxy.yml file around tool filtering. Simon's commits from when he introduced subdomains to dev last year might contain these.

neoformit:
Damn, yeah you're right. I'll check that out thanks.

neoformit:
Hey @cat-bro just removed/commented out for AARNet and added to staging. I figured we should really be testing the playbook there first. There were some certbot issues when we deployed this on dev, I [documented it in the ops manual](https://github.com/usegalaxy-au/operations/pull/11). Based on Simon's commits and the my notes I think this is all the conf required for staging.

cat-bro (approved):

PR #1329, opened 2023-06-20
List gx_it_proxy as a service on dev, staging and reduce number of job handlers on staging

cat-bro:
It seems that the gie-proxy role is no longer required and gravity will start this up as a service.


nuwang (approved):
LGTM!

PR #1333, opened 2023-06-22
Deploy genomes lab to AARNet

neoformit:
This PR will deploy the "genomes lab" subdomain, the `webhooks` role and the `subdomain_switcher` webhook (masthead icon) to aarnet/prod. It also applies some updates to these roles on dev and staging.

Run the aarnet playbook with `--extra-vars "certbot_expand=true"` to force certbot to create a new certificate that includes the `genome.usegalaxy.org` domain.

The webhooks role will notify `restart galaxy` - please make sure that galaxy restarts for the `subdomain_switcher` webhook to take effect. It should look like this (live on staging):

image

### DIFF
--- a/aarnet_playbook.yml
+++ b/aarnet_playbook.yml
@@ -45,7 +45,7 @@
       become: true
       become_user: galaxy
     # - usegalaxy_eu.galaxy_systemd
-    # - usegalaxy_eu.galaxy_subdomains
+    - usegalaxy_eu.galaxy_subdomains
     - nginx-upload-module
     - galaxyproject.nginx
     - galaxyproject.tusd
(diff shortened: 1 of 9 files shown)


cat-bro:
Do we need to use to force option when adding a new domain?

neoformit:
> Do we need to use to force option when adding a new domain?

Yes, it skips the certbot task if a config file already exists - see [ops manual](https://github.com/usegalaxy-au/operations/blob/main/head_node/subdomains.md#ssl-certificates). So when you add a new value to `certbot_domains` you either have to use the `certbot_expand` var or re-issue manually.

cat-bro (approved):

PR #1341, opened 2023-06-23
Add activation template dev

cat-bro:
Activation email for dev. I copied Cameron's template from the galaxy codebase https://github.com/galaxyproject/galaxy/blob/release_23.0/lib/galaxy/config/templates/mail/activation-email.html and changed it to GA's email from the google doc. There is a commit showing the difference between one and the other.

@neoformit does there need to be a plain text version of this as well or is the html enough?


neoformit:
Good to have a plain text version too. Email clients that can't do HTML (if they still exist) will render the plain text version. I think if you omit the text template it will fall back on the default template. Worth checking this after deploying to dev!

neoformit (commented):
Looks good but probably include text email template before deploying to prod

neoformit (approved):

PR #1345, opened 2023-06-26
don't abort in galaxy-backup-db.sh script when 'swift delete' fails

jlqfab:


cat-bro:
@jlqfab this looks great. There is at least one discrepancy between values set for the ansible role and values in the bash script on aarnet-backup so it's a good idea to check these before merging this.

jlqfab:
@cat-bro  Diff on existing script on aarnet-backup and generated script from template has been done. Changes have been made to template to add the addition swift options currently used.

cat-bro (approved):

PR #1356, opened 2023-07-02
Deploy proteomics lab

neoformit:
Requires [CNAME DNS record](https://github.com/usegalaxy-au/operations/blob/main/head_node/subdomains.md)
and run `ansible-playbook` with `--extra-vars "certbot_expand=true"` to force new SSL cert generation.

See also https://github.com/usegalaxy-au/galaxy-media-site/pull/57

### DIFF
--- a/host_vars/aarnet.usegalaxy.org.au.yml
+++ b/host_vars/aarnet.usegalaxy.org.au.yml
@@ -631,6 +631,30 @@ galaxy_subsites:
       #masthead .navbar-text {
         display: none;
       }
+  - name: proteomics
+    brand: Proteomics lab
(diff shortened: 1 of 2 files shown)


cat-bro:
LGTM! The `tool_section_filters` in host_galaxy_config.galaxy will need to be uncommented in host_vars/aarnet for the section filtering to work on genome and proteomics sites. Do you want to include `proteomic_ai` in this one too for the alphafold users?

neoformit:
We decided not to do toolbox filtering in the end so technically fine as it is. We should probably remove the tool filter sections to reflect that choice though.

cat-bro (approved):

PR #1363, opened 2023-07-04
Fix mail template

neoformit:
Remove broken/redundant gfonts links. Email clients can't use these and it affects spam score.


cat-bro:
Thanks @neoformit!

PR #1366, opened 2023-07-06
add safery checks to attached-volumes role

jlqfab:


roles/attached-volumes/tasks/set_up_volume.yml
```diff
@@ -1,27 +1,65 @@
...
-  when: attached_volume.device is defined and attached_volume.fstype is defined and attached_volume.path is defined
+  - name: "Detect attached volume"
+    parted:
+      device: "{{ attached_volume.device }}"
+      unit: GiB
+    register: volume_info
+
+  - name: "Set fact for number of existing partitions"
+    set_fact:
+      num_parts: "{{ volume_info.partitions | length }}"
```
cat-bro, line 11: Is there a way to make this an int at this step, same with num_parts2

cat-bro (approved):

PR #1394, opened 2023-07-26
Etca telegraf

jlqfab:
Define influx database galaxy_etca - to be created by running the stats_playbook.yml.
Enable telegraf on etca playbooks (requires stats_playbook.yml to be run beforehand).


cat-bro (approved):
LGTM!

PR #1413, opened 2023-08-04
Add ssh key for accessing workers from galaxy-backup, galaxy, galaxy-queue

cat-bro:
From galaxy-queue we could ssh to a worker with 'ssh galaxy-w10' for example. galaxy-w10 will only accept this ssh key from galaxy, galaxy-queue or galaxy-backup.

This does not allow every galaxy machine access to every other one - nothing can use this key to get to the galaxy VM and we should probably allow access from galaxy-backup, in which case this could actually replace galaxy-australia as the deployment key.


jlqfab (dismissed):
does galaxy-handlers need to be added somewhere?

cat-bro:
this isn't ready yet, just rebased

cat-bro:
@jlqfab this is ready for review

group_vars/galaxy_workers.yml
```diff
@@ -28,3 +28,9 @@ docker_users:
   - "{{ galaxy_user.name }}"
 docker_daemon_options: # TODO: Uncomment as soon as there is a volume
   data-root: /mnt/docker-data
+
+# internal ssh keys
+extra_keys:
+  - id: internal_hop_key
+    type: public
+    from: "{{ hostvars['galaxy']['internal_ip'] }},{{ hostvars['galaxy-backup']['internal_ip'] }},{{ hostvars['galaxy-queue']['internal_ip'] }}"
```
jlqfab, line 36: should we also add galaxy-handlers internal_ip here?

cat-bro: yes. pushing change now.

roles/common/tasks/main.yml
```diff
@@ -91,6 +90,14 @@
   loop_control:
     loop_var: extra_key
 
+- name: Add ssh config
+  import_tasks: ssh_config.yml
+  when: ssh_config_hosts is defined
+
+- name: Add ssh config
```
jlqfab, line 97: this duplicates the above task

host_vars/galaxy-handlers.usegalaxy.org.au.yml
jlqfab, line 1: I think this also needs the following defined, as per galaxy, galaxy-backup, galaxy-queue
```
# ssh config, only for ubuntu
ssh_config_id_file: "/home/{{ ssh_config_user }}/.ssh/internal_hop_key"
ssh_config_user: ubuntu
ssh_config_hosts: "{{ galaxy_group }}"
```

roles/common/tasks/main.yml
cat-bro: oops. new commit removes the second copy

host_vars/galaxy-handlers.usegalaxy.org.au.yml
cat-bro: it will actually inherit that from galaxy.usegalaxy.org.au

jlqfab (approved):

PR #1423, opened 2023-08-11
Galaxy slurm update and memory corrections

cat-bro:
Based on Tim Wickberg's advice [here](https://bugs.schedmd.com/show_bug.cgi?id=3415) and yesterday's experience, set RealMemory for all slurm workers to something a little less than MemTotal. For the ordinary workers that can be a very round number: 32000, 64000 or 128000. The values have also been lowered a little for the high-memory and training nodes. They will still all have enough stated memory for the jobs being sent to them.

### DIFF
--- a/group_vars/aarnet_slurm.yml
+++ b/group_vars/aarnet_slurm.yml
@@ -11,52 +11,52 @@ slurm_nodes:
   - name: aarnet-w1
     NodeAddr: "{{ hostvars['aarnet-w1']['internal_ip'] }}"
     CPUs: 16
-    RealMemory: 62815
+    RealMemory: 64000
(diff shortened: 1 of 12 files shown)


jlqfab:
in group_vars/pulsar_mel3/pulsar-mel3_slurm.yml

some values are larger than before
-    RealMemory: 125812
+    RealMemory: 128000
this is for some of the pulsar-mel3-w* nodes

cat-bro:
yes, some of the values are larger than before. The values being reported back yesterday were exactly the MemTotal values of the VMs in question, so I am applying this to all of them.

jlqfab (approved):

PR #1424, opened 2023-08-11
Add temporary rule for jenkins_bot so that msconvert tests are avoided

cat-bro:


cat-bro:
This was an attempted workaround to stop a disk filling up in middle of the night. It failed linting (not sure why) so I couldn't merge it and the disk filled up. It's not a great solution anyway, gonna close this

PR #1429, opened 2023-08-16
Disable UFW on all VMs until allow rules are configured

jlqfab:


roles/common/tasks/main.yml
```diff
@@ -38,6 +38,26 @@
...
+- name: Configure creation of user home directories to be user group readable only
+  lineinfile:
+    dest: "/etc/adduser.conf"
+    regexp: "^DIR_MODE="
+    line: "DIR_MODE=0750"
+    state: present
+
+- name: Ensure the system user's home directory is private
+  file:
+    path: "/home/{{ ansible_user }}"
```
cat-bro, line 57: what is the purpose of this task? ubuntu won't always be the ansible user. When playbooks are run from jenkins tasks the ansible user is jenkins_bot.

cat-bro:
This looks good but I would prefer to do this as a one-off than add these tasks to the common role.

roles/common/tasks/main.yml
jlqfab: This is just to ensure the system user (ubuntu) home directory is private as per AARNet's security concerns. I didn't realise the issue with jenkins_bot. I can change it to /home/ubuntu

jlqfab:
@cat-bro I can put all these tasks in an etca.yml task within the common role. The disabling of Apport is a separate issue, but will prevent some of our core dumps. It may need to go into a workers playbook.

cat-bro:
I agree. Keep the disabling of apport in the common role

roles/common/tasks/main.yml
cat-bro: Does aarnet want the mode of /home/ubuntu to be 0700?

jlqfab: That is my memory of the discussion with Ollie. Their whole rationale for making everyone's home directories private was because of access to the system user (ubuntu) home directory being open by default. Compromise is that we make it private, mitigating some of their security concerns.

cat-bro: I can't see anything in /home/ubuntu on aarnet that would need to be r'd or x'd by anyone other than ubuntu.

cat-bro (approved):

nuwang:
Did we have a list of ports that were required? I have this recollection that it was compiled. Just for future ref.

Issue #1427, opened 2023-08-15
blast jobs running with stored reference data can only go to pulsar-qld-blast


cat-bro:
The pulsar-qld-blast VM is optimised for blast jobs with an nvme volume. It has 60 CPUs and each job is granted 15. Blast jobs can only run on slurm unless they are using reference data from cvmfs or custom-indices, in which case they will go to pulsar-qld-blast. The way this has been set up in TPV, there is no alternative destination for these jobs when the VM is at capacity and jobs will queue or stay in new state.

There needs to be overflow of these jobs to another pulsar destination, probably pulsar-mel3, with a ranking bias to always send a job to pulsar-qld-blast if there is capacity.

PR #1430, opened 2023-08-16
Allow pulsar blast jobs to go to pulsar-mel3

cat-bro:
To alleviate the blast queue issue.

What we also need potentially is an alternative ranking function for these tools that will send jobs to a second destination *only* if the primary destination is full.

This is related to issue https://github.com/usegalaxy-au/infrastructure/issues/1427 and would be good to discuss with @nuwang and @bgruening


cat-bro:
I think we need to merge this for the currently queued jobs

jlqfab (approved):

PR #1432, opened 2023-08-16
Remove score from consideration in TPV ranking

cat-bro:
Currently a destination with the highest score will always be returned first in ranking, even if it 100% occupied.


nuwang (approved):
Are we ok with completely disregarding the score? because once a destination matches at some basic level, we will no longer be taking preferences into account. A stronger match would be ignored in favour of a weaker match if the cpu usage is lower in the weaker match. Perhaps the score should be calculated differently?

Perhaps we can calculate some normalized multiplication of the cpu and score?
For now though, makes sense to simply ignore score.

cat-bro:
The way the score is calculated seems good. The problem is that the sorting is done in such a way that score eclipses any other measure. This is the reason we don't use prefer very often, as it is effectively the same as require if we rank that way.

PR #1441, opened 2023-08-22
Limit clustalw to two jobs per person

cat-bro:


cat-bro:
Someone is currently queueing 6 of these. Merging this so that they can only queue two at a time.

PR #1446, opened 2023-08-24
add Cloudflare API token for GVL domain and specify this token in hos…

jlqfab:
…t_vars/dev.gvl.org.au.yml


jlqfab:
added cloudflare API token for GVL domain to group_vars/VAULT and updated host_vars/dev.gvl.org.au.yml to use it.

host_vars/dev.gvl.org.au.yml
```diff
@@ -32,7 +32,7 @@ certbot_domains:
   - "genome.{{ hostname }}"
 certbot_dns_provider: cloudflare
 certbot_dns_credentials:
-  api_token: "{{ vault_dns_cloudflare_api_token }}"
+  api_token: "{{ vault_dns_cloudflare_gvl_api_token }}"
```
cat-bro, line 35: does this token need to be used for dev-queue too?

cat-bro (approved):

host_vars/dev.gvl.org.au.yml
jlqfab: I don't think so - it should only be needed where `certbot_dns_credentials` is defined, which for the dev vms is only `host_vars/dev.gvl.org.au.yml`

PR #1459, opened 2023-08-28
Distributed object store conf for aarnet

cat-bro:
This can't be merged yet. Still to do:

(1) Set object store IDs for datasets and metadata files for all 'store_by="id"' backends (so far this has only been set for perthNFS1-4)
(2) Update aarnet-user-nfs playbook to create new folder `data008` alongside user-data5 and user-data7 in /mnt directory on aarnet-user-nfs and a new subdirectory `data009` of /user-data6
(3) Update aarnet and aarnet_workers shared_mounts variables to mount the new folder /mnt/data008

To switch to using this configuration:
Stop galaxy.
Set all dataset ids and metadata_file object_store_ids created since August to aarnetNFS7
Run aarnet_update_configs playbook
Start galaxy

### DIFF
--- a/templates/galaxy/config/aarnet_object_store_conf.xml.j2
+++ b/templates/galaxy/config/aarnet_object_store_conf.xml.j2
@@ -1,34 +1,43 @@
 <?xml version="1.0"?>
-<object_store type="hierarchical">
+<object_store type="distributed" id="primary" order="0">
     <backends>
-        <backend id="aarnetNFS7" type="disk" order="0">
+        <!-- New folder with storage by UUID on the 50T volume. Initially set weight=0. -->
+        <backend id="data09" weight="0" store_by="uuid">
+            <files_dir path="/mnt/user-data-6/data09" />
+        </backend>
+        <!-- New folder with storage by UUID on the 150T volume, weight=1 means it is the only
+        one that will be written to if all of the others have weight=0 -->
+        <backend id="data08" weight="1" store_by="uuid">
+            <files_dir path="/mnt/data08" />
+        </backend>
+        <backend id="aarnetNFS7" type="disk" order="0" store_by="id">
             <files_dir path="/mnt/user-data-7" />
         </backend>
-        <backend id="aarnetNFS6" type="disk" order="1">
+        <backend id="aarnetNFS6" type="disk" weight="0" store_by="id">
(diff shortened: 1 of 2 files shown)


cat-bro:
@jlqfab I was thinking of using terse naming for object store folders and IDs starting from 008 (I think for legacy reasons we need to live with the fact that both qldNFS1 and perthNFS1 exist)

jlqfab (commented):
`        <backend id="aarnetNFS7" type="disk" order="0">
        
        <backend id="data009" weight="0" store_by="uuid">
            <files_dir path="/mnt/user-data-6/data009" />
        </backend>
        
        <backend id="data008" weight="1" store_by="uuid">
            <files_dir path="/mnt/data008" />
        </backend>`
So the intention is that /mnt/data008 is the one written to? Ie on the 150T volume. The comment on data009 is a bit confusing. "... b <newline> one that will be written to if all others have weight 0..." is this something to do with user-data-6/6b ?

templates/galaxy/config/aarnet_object_store_conf.xml.j2
```diff
@@ -1,34 +1,44 @@
...
+        one that will be written to if all of the others have weight 0  -->
+        <backend id="data009" weight="0" store_by="uuid">
+            <files_dir path="/mnt/user-data-6/data009" />
+        </backend>
+        <!-- New folder with storage by UUID on the 150T volume, weight=1 means it is the only
+        one that will be written to if all of the others have weight=0 -->
+        <backend id="data008" weight="1" store_by="uuid">
+            <files_dir path="/mnt/data008" />
+        </backend>
+        <backend id="aarnetNFS7" type="disk" order="0" store_by="id">
```
jlqfab, line 14: So the intention is that /mnt/data008 is the one written to? Ie on the 150T volume. The comment on data009 is a bit confusing. "... b <newline> one that will be written to if all others have weight 0..." is this something to do with user-data-6/6b ?

cat-bro: the second line of comment on 009 has been pasted in error

jlqfab: thanks. that makes more sense.

jlqfab (approved):

PR #1461, opened 2023-08-29
mem for funannotate_clean

TomHarrop:
Not sure how high to go for this tool. It does an all vs all alignment of contigs so it could potentially use a lot of memory


cat-bro:
@TomHarrop can you try setting this to 'mem: 10'? I think this tool might not need 30 and jobs might queue for longer than they need to with this setting.

cat-bro (approved):

PR #1464, opened 2023-08-31
add oom query to gxadmin local

cat-bro:
`gxadmin local query-oom-jobs` lists the jobs with `This job was terminated because it used more memory than it was allocated.` in job.info or 'Killed' in job.tool_stderr.


jlqfab (approved):
Nice. But curious about needing to test for both 'Killed' in job.tool_stderr (I assume this is where a cgroups killed a job) and the explicit out of memory error in job.info (slurm reported error?).

PR #1500, opened 2023-09-12
run mothur with docker for safety

cat-bro:
mothur jobs will fail if any file in the working directory is bigger than 470G. It will also fail if too many pids are created.

[EDIT]: I previously wrote that this would not work if the  tool ID was not in our local config - have since tested this with dry run and it should work either way.


jlqfab (approved):
for noting: ulimit file size is 476.8G

PR #1502, opened 2023-09-12
define hosts_template_file for galaxy_etca

jlqfab:
Fix cloud init config issue causing /etc/hosts to be wiped on reboot, by defining `hosts_template_file` as `/etc/cloud/templates/hosts.debian.tmpl` for galaxy etca.

### DIFF
--- a/group_vars/galaxy_etca.yml
+++ b/group_vars/galaxy_etca.yml
@@ -3,6 +3,9 @@ galaxy_gid: 10010
 
 use_internal_ips: true
 
+# cloud-init template for /etc/hosts
+hosts_template_file: "/etc/cloud/templates/hosts.debian.tmpl"
+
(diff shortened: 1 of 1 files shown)


cat-bro:
I was under the impression this would not be happening at etca. I have recently rebooted galaxy-handlers and it did not lose any hosts.

jlqfab:
I saw the `update_etc_hosts` setting in `/etc/cloud/cloud.cfg` on galaxy.usegalaxyy.org.au. This is the setting that caused issues with aarnet.usegalaxy.org.au. What has changed in the config?

jlqfab:
Olivier Allart:
> that issue #1502 - cloud init is fixed
> there is NO provision in cloud init in ETCA to manage etc-hosts

PR #1506, opened 2023-09-13
add fgenesh and cellranger to tools.yml

mthang:


cat-bro:
@mthang I had though that there was a rule like this for maxquant_test but there isn't. The only access rules we have so far are for alphafold. I'm not sure if the names 'cellranger' and 'fgenesh' (for the role names) might be too informal - I'll write on slack.

cat-bro:
@mthang this still needs to have the params section with `singularity_enabled: true`

This needs to NOT have the singularity_volumes line.  For both tools.

cat-bro:
closing in favour of the other one

PR #1531, opened 2023-09-23
fix galaxy job conf

cat-bro:


cat-bro:
pulsar-paw was trying to post files back to galaxy-handlers.usegalaxy.org.au, needs to be galaxy.usegalaxy

PR #1538, opened 2023-09-27
Enable subdomain cookie-sharing on prod

neoformit:
Not sure if we should merge this without testing first. But I'm not sure how to test this?

### DIFF
--- a/host_vars/aarnet.usegalaxy.org.au.yml
+++ b/host_vars/aarnet.usegalaxy.org.au.yml
@@ -226,6 +226,9 @@ host_galaxy_config: # renamed from __galaxy_config
     # tool_section_filters: global_host_filters:per_host_tool_sections
     # tool_label_filters: global_host_filters:per_host_tool_labels
 
+    # Allow cross-subdomain cookie sharing:
+    cookie_domain: usegalaxy.org.au
+
     user_activation_on: true
     activation_email: <activation-noreply@usegalaxy.org.au>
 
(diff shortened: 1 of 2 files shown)


cat-bro:
from galaxy.yml.sample:
```
  # Tell Galaxy that multiple domains sharing the same root are
  # associated to this instance and wants to share the same session
  # cookie. This allow a user to stay logged in when passing from one
  # subdomain to the other. This root domain will be written in the
  # unique session cookie shared by all subdomains.
  #cookie_domain: null
```

I feel like we can wing it. The only other galaxy now using usegalaxy.org.au is Sydney (galaxy.usegalaxy.org.au) and if cookie_domain is still null on that instance, it shouldn't be affected (I think). If we add this (commented out) to host_vars/galaxy.usegalaxy.org.au that is something we don't have to remember later on.

neoformit:
I'm not sure about that - I think if `usegalaxy.org.au` sets a cookie for `usegalaxy.org.au` then that cookie can be accessed by all subdomains unless there is some internal Galaxy logic that decides to ignore it. But I guess we can test it out - just see what happens when you access `galaxy.usegalaxy.org.au` while logged into `usegalaxy.org.au`. I feel like they might have to share a database in order to even find the user that's supposed to be logged in?

cat-bro (approved):

jlqfab (approved):

cat-bro:
Thanks @neoformit !

PR #1545, opened 2023-10-03
Add opts to roles that mount volumes

cat-bro:
Add 'opts' to ansible mount tasks.

So far we've been dealing 2-VM galaxy issue by setting actimeo=2 for the user data mount on the galaxy VM. This is not the only thing we could do here - we could be using noac and there are probably other options as well. I'm still curious as to how the current deployment doesn't have issues caused by nfs cache time, but then maybe it does, and some of the nfs/io bottlenecks are due to this.


jlqfab (approved):
Looks good. For later reference - as earlier noted `noac` option is not available on NFSv4.

PR #1570, opened 2023-10-18
remove galaxy etca from influxdb 14d retention policy

jlqfab:
galaxy_etca influx database has manually had a "forever" retention policy applied to it, to keep monitoring of etca tests. This change prevents the usual default retention of 14 days being applied when the stats playbook is run. When this data is no longer required, this change can be reverted and the retention policy changed to the standard 14 days.


nuwang:
It would be really nice if we could use some of these downsampling features: https://www.influxdata.com/blog/downsampling-influxdb-v2-0/
Would probably be portable to other Galaxy instances as well.

cat-bro (approved):

PR #1604, opened 2023-10-29
Add weight for slurm nodes

cat-bro:


cat-bro:
At the moment all of the jobs on the local slurm cluster are running on the 32 core nodes. This should cause the smaller nodes to be prioritised so that the larger nodes are available for larger jobs.

PR #1616, opened 2023-11-03
Format tpv config and add more singularity tools

cat-bro:
Add singularity_enabled for a number of tools have difficult conda environments. I've also run 'tpv format' on tools.yml which resulted in some lost comments that did not seem worth preserving.


cat-bro:
Formatting was necessary in order to add the tools programmatically. It sorts they keys alphabetically. I'd prefer to keep comments in future - this is something I will follow up with TPV.

jlqfab (approved):

PR #1624, opened 2023-11-12
replace pawsey with galaxy etca for maintenance key on pulsar-QLD-job-nfs

jlqfab:


cat-bro (commented):
this is not used. could you please remove the extra_keys block instead?

cat-bro (approved):

PR #1644, opened 2023-11-22
let phm2 test user run fgenesh on phm2

cat-bro:
the pulsar test users are set up to run only toolshed tools on the pulsars and all other jobs go to local slurm. phm2 test user needs a workaround so that fgenesh it runs go to phm2.


jlqfab:
Any reason you're removing the high-mem spec for the other high memory pulsars for jobs run by the test user?

jlqfab:
cat-bro: this tag is no longer used for scheduling, it is redundant

jlqfab (approved):

PR #1658, opened 2023-12-04
Add proteomics panel view

neoformit:
Ok, so this PR doesn't have that WIP nonsense that broke prod last time.
A new config file referenced by dev host vars only.

### DIFF
--- a/host_vars/dev.gvl.org.au.yml
+++ b/host_vars/dev.gvl.org.au.yml
@@ -100,6 +100,8 @@ host_galaxy_config_files:
     dest: "{{ galaxy_config_dir }}/mail/activation-email.txt"
   - src: "{{ galaxy_config_file_src_dir }}/config/local_tool_data_table_conf.xml"
     dest: "{{ galaxy_config_dir }}/local_tool_data_table_conf.xml"
+  - src: "{{ galaxy_config_file_src_dir }}/config/tool_panel_filters/proteomics.yml"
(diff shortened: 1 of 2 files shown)


cat-bro:
I think this needs https://github.com/galaxyproject/ansible-galaxy/blob/main/defaults/main.yml#L125C1-L125C30

neoformit:
Looking at galaxyserver.yml - looks like I should be using this var and not setting my own
https://github.com/usegalaxy-au/infrastructure/blob/8e7196f2e0ed1f6b70ae23f802923041e3874f84/group_vars/galaxyservers.yml#L127

But I can add `"{{ galaxy_config_dir }}/mail"` to fix your other issue.

cat-bro:
Thanks, that closes out an issue from a while ago

host_vars/dev.gvl.org.au.yml
```diff
@@ -100,6 +100,8 @@ host_galaxy_config_files:
     dest: "{{ galaxy_config_dir }}/mail/activation-email.txt"
   - src: "{{ galaxy_config_file_src_dir }}/config/local_tool_data_table_conf.xml"
     dest: "{{ galaxy_config_dir }}/local_tool_data_table_conf.xml"
+  - src: "{{ galaxy_tool_box_filters_dir }}/proteomics.yml"
```
cat-bro, line 103: this is not going to work. src should be w.r.t the infrastructure directory - galaxy_tool_box_filters_dir is on the destination VM

PR #1659, opened 2023-12-04
more medium minimap memory

cat-bro:
there is probably a better solution, like scaling memory to input size, but this might do for now


cat-bro:
this might be bogus

Issue #1660, opened 2023-12-05
External file mounts refactor


cat-bro:
There are boolean flags for easily removing external file mount paths from the object store conf, and from singularity_volumes strings but this other problems to solve:

(1) File paths such as /mnt/user-data, which are mount points for pawsey-user-nfs volumes, will be bound to existing singularity jobs through the singularity_volumes value when the jobs began, and this can cause jobs to become stuck.

(2) If mounts are offline, the 'mounts' role will go through the difficult mounts and get stuck, so I've been manually commenting the mounts role out of the galaxy_playbook.

For (2), we can use conditionals to choose the state based on whether the mounts are available, e.g.
```
  - path: /mnt/user-data
    src: "pawsey-user-nfs.usegalaxy.org.au:/mnt/user-data"
    fstype: nfs
    opts: 'noatime,defaults'
    state: "{{ 'mounted' if pawsey_file_mounts_available else 'absent_from_fstab' }}"
```
The only sensible state options for an unavailable mount are 'present' (do nothing) or 'absent_from_fstab' (remove from fstab and do nothing else). Anything involving unmounting these with ansible will almost certainly run into errors, as manual unmounting does.

For (1), it seems like a good idea to move the mounts points into directories, so that no mount point is itself a path in singularity_volumes (there will be /mnt/user-data-pawsey/user-data, /mnt/user-data-pawsey/user-data-2 etc) then the entry in singularity_volumes is /mnt/user-data-pawsey, a normal directory path that will not cause problems.

This cannot be done in one PR - I'm tracking it as an issue to make a checklist and keep track of it.

cat-bro:
- [ ] Create containing folders for pawsey and qld file mounts on galaxy, galaxy-handlers and workers
- [ ] Run playbooks when PR 1661 is merged.
- [ ] Edit shared_mounts entries for pawsey so mounted things are in /mnt/user-data-pawsey
- [ ] Add additional shared_mounts entries for qld so mounted things are ing /mnt/user-data-qld (doubled up so not to harm running jobs)
- [ ] Run playbooks again
- [ ] Update singularity_volumes list in host_vars/galaxy and object store conf
- [ ] Run galaxy playbook and restart galaxy-handlers

PR #1663, opened 2023-12-05
volume refactor part 2

cat-bro:
Update pawsey mount points so that when they are remounted they are within /mnt/user-data-pawsey. Update object store conf for new paths, and singularity_volumes.


cat-bro:
following checklist items here https://github.com/usegalaxy-au/infrastructure/issues/1660

jlqfab (approved):
looks good!

PR #1677, opened 2023-12-08
fix path for proteomics.yml on dev

cat-bro:
Fix path for file to copy for proteomics tool panel filtering

@neoformit you were saying this change did not work as expected the other day - maybe the playbook was not rebased.


neoformit:
Thanks Cat I did change this locally - and went back to using the `panel_views_dir`. Probably better to close this and I'll open a new one. I got something working in the end...

PR #1696, opened 2023-12-20
move SACCT_FORMAT for root user to common role

cat-bro:
`sudo sacct` is available to all machine users, not just sudo users, on any VM with slurm installed (galaxy VM, worker nodes, all pulsar machines etc) but the environment variable setting the output to a useful format format has only been set on slurmcontroller VMs and has been missing from workers and galaxy web/handlers servers.


nuwang (approved):
lgtm!

PR #1700, opened 2023-12-22
Remove pg-post-tasks and add common bash variable and functions

cat-bro:
Remove time-consuming pg-post-tasks role and put its tasks into the common role in a more efficient form. If there are environment variables which need to be the same for all users, these can be set in one common file.

Add bash functions for all users `jwd`, `go-jwd` and others. We can add more of these and ones that are helpful on pulsar VMS, useful aliases etc over time.

Add a tag `users` to the tasks that loop over all the users so that they can be easily skipped.


cat-bro:
Hi @jlqfab, we can add other common bash variables here as well like `export EDITOR=vim`. This is not ready for merge because I'd like to add a few more things such as a `jwd` function for pulsar

jlqfab (approved):
This looks good!

jlqfab:
@cat-bro besides the `EDITOR` export, I like to change the defaults in the `.bashrc` for `HISTSIZE` and `HISTFILESIZE`, so that history doesn't get erased too quickly (1000 lines of history is exceeded very quickly).

cat-bro:
The perfect is the enemy of the decent, this is probably good to go

hosts
```diff
@@ -146,15 +146,30 @@ pulsar_qld_blast
...
+# All pulsar head nodes
+[pulsar_head_nodes:children]
+production_pulsars
+dev_pulsar_head
+staging_pulsar_head
+pulsar_nci_test_head
+
+# All pulsar workers
+[pulsar_workers:children]
+pulsar_paw_workers
```
jlqfab, line 158: Isn't pulsar-paw gone?

jlqfab (approved):
Looks good!

hosts
cat-bro: Thanks JL! I need to add pulsar_qld_gpus to production_pulsars as well

jlqfab (approved):

PR #1727, opened 2024-01-19
slurmdbd refactor

cat-bro:
The slurmdbd configs for phm1 and phm2 were slightly wrong, resulting in no slurmdbd log files. It turns out that all of the VMs using slurmdbd can use the same config block so this is moved to all.yml along with MariaDB config which is the same everywhere.

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -279,3 +279,33 @@ __slurmdbd_config_default: # this is not needed if this PR is merged: https://gi
   DbdPort: 6819
   SlurmUser: "{{ __slurm_user_name }}"
   LogFile: "{{ __slurm_log_dir ~ '/slurmdbd.log' if __slurm_debian else omit }}"
+
+slurmdbd_config:
+  AuthType: auth/munge
+  AuthInfo: /var/run/munge/munge.socket.2
+  DbdHost: localhost
+  StorageHost: localhost
+  StorageLoc: "{{ slurm_database }}"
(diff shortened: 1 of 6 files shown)


group_vars/all.yml
```diff
@@ -279,3 +279,33 @@ __slurmdbd_config_default: # this is not needed if this PR is merged: https://gi
...
+  AuthInfo: /var/run/munge/munge.socket.2
+  DbdHost: localhost
+  StorageHost: localhost
+  StorageLoc: "{{ slurm_database }}"
+  StoragePass: "{{ slurm_database_user_password }}"
+  StorageType: accounting_storage/mysql
+  StorageUser: "{{ slurm_database_user }}"
+  PidFile: /run/slurmdbd.pid
+  SlurmUser: slurm
+
```
jlqfab, line 294: Does `LogFile` need to be defined here?

cat-bro: No, because the value from `__slurmdbd_config_default` will be used

jlqfab (approved):

cat-bro:
To add context: the slurm version on ubuntu 22.04 is higher than that on 20.04 and the config and log paths are /etc/slurm and /var/log/slurm instead of the previous /etc/slurm-llnl and /var/log/slurmllnl. This is why phm1 and phm2, who inherited slurmdbd config from pulsarservers.yml, ended up with wrong paths in the config. Luckily (thanks to Nate) the slurm role takes care of all of these differences.

PR #1730, opened 2024-01-19
delete aarnet playbooks and files

cat-bro:


jlqfab (commented):
Note that there are also the aarnet files under `templates/galaxy/config/`: `aarnet_job_conf.yml.j2` and `aarnet_object_store_conf.xml.j2` which I assume are superseded by the `galaxy_*` files in same directory.

jlqfab (dismissed):

jlqfab (approved):

PR #1733, opened 2024-01-22
Clean tmpdisk role

cat-bro:
Python bash combination for finding how long the longest running job has been running and removing tmp data with larger mtime.

### DIFF
--- a/galaxy-workers_playbook.yml
+++ b/galaxy-workers_playbook.yml
@@ -31,6 +31,7 @@
       - usegalaxy_eu.apptainer
       - geerlingguy.docker
       - acl-on-startup
+      - clean-tmpdisk
   post_tasks:
       - name: restart munge
         systemd:
--- a/group_vars/pulsar_paw/pulsar-paw_workers.yml
+++ b/group_vars/pulsar_paw/pulsar-paw_workers.yml
@@ -2,6 +2,8 @@ slurm_roles: ['exec']
 
 add_hosts_head: yes
 
+clean_tmpdisk_tmp_dir: /tmp  # no mounted tmpdisk
+
 shared_mounts:
     - path: /mnt/custom-indices
       src: pulsar-paw:/mnt/custom-indices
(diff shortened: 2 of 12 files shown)


nuwang (commented):
This is a nice idea! One suggestion - once the oldest job is determined - would it be better to use `tmpwatch` to cleanup here, since it handles more edge cases etc?

cat-bro:
Does tmpwatch handle more edge cases? Doesn't it need to be set with a flat time such as 'clean up files older than 30 days'?

cat-bro:
yes you're right. I think it's tmpreaper for debian https://manpages.ubuntu.com/manpages/xenial/man8/tmpreaper.8.html

jlqfab:
Is there any real need to use `tmpreaper` given the issue with it creating a default cronjob (that could be removed) when the manpage (linked above) says

> Unless  your  machine  is  one  with  lots  of  relatively untrusted users, such as an ISP or school, you don't need this program; `find ... -exec rm ...' works just as well when you don't have to be concerned about people trying to exploit the race condition on you.

cat-bro:
Thanks @jlqfab . What I like about tmpreaper is that it takes away the need for me to account for options like test/dry-run and verbose, but what I dislike about it is the automatic cron job (which can be removed, but I don't like having to rely a removal task that could fail). If it's deprecated in favour of systemd-tmpfiles I lose some of the nice features, so I might go back one commit to the original implementation. For my record this is commit 7d38f03 in case I want to cherry-pick it.

jlqfab (approved):

nuwang:
@cat-bro Can this become a gxadmin command? This would be super useful for the Galaxy helm chart as well.

cat-bro:
It doesn't fit with gxadmin because it's only run on worker VMs that are not connected to the galaxy database. I'm working on scripts for running slurm commands remotely via ssh (such as squeue, sacct) which could potentially fit into gxadmin - it would be good to have some integration of the the info from the slurm and galaxy databases. I'm not sure how to make this one more accessible - the rm commands can take a long time to complete so I wouldn't want to be calling them via ssh.

nuwang:
That makes sense - not much choice in that case.

PR #1736, opened 2024-01-23
Add stats script to add volume info to GA_server database

cat-bro:


jlqfab (approved):
Nice.

nuwang:
An alternative possibility here is to use the [Prometheus node exporter](https://github.com/prometheus/node_exporter). It can be run as a docker container which scrapes all relevant information (detailed cpu, memory, disk stats etc.). This is a dashboard with some of that information: https://grafana.com/grafana/dashboards/1860-node-exporter-full/.

jlqfab:
I use prometheus node exporter on Apollo. But Galaxy is using influx not prometheus for its time series database.

nuwang:
Use telegraph to adapt? https://docs.influxdata.com/influxdb/cloud/write-data/developer-tools/scrape-prometheus-metrics

cat-bro:
We are using telegraf. We have disk info data points stored once per minute in a db with a 2-week retention policy. The aim here is to collect sparse data that we never delete. It was pretty quick to pull this together - my preference would be to merge it as is and drop the measurement later if we have another solution. Ultimately it would be nice to retain all data at sparse points after the initial two weeks of saturation.

nuwang:
Wouldn't it be better to downsample the data? https://www.influxdata.com/blog/downsampling-influxdb-v2-0/

The dashboards are fairly mature, and carry a lot of information that's useful, so we could avoid some maintenance costs etc.

No objections to merging this as is though, this can be a future task if there's sufficient need.

cat-bro:
> Wouldn't it be better to downsample the data?

This looks great! Would save us from kicking can down the road in terms of other stats. I'm going to look into it.

PR #1769, opened 2024-02-08
new interactive tool urls

cat-bro:
Since updating dev to release_23.2 interactive tools are being mapped to < some uuid >.ep.interactivetool.dev.gvl.org.au instead of < some uuid >.interactivetoolentrypoint.interactivetool.dev.gvl.org.au.  After creating a new CNAME record for dev and updated the certificate to include the new pattern, ITs are working on dev. I could be missing something about how ITs are supposed to work post 23.2.

### DIFF
--- a/host_vars/dev.gvl.org.au.yml
+++ b/host_vars/dev.gvl.org.au.yml
@@ -25,6 +25,7 @@ shared_mounts:
 certbot_domains:
   - "{{ hostname }}"
   - "*.interactivetoolentrypoint.interactivetool.{{ hostname }}"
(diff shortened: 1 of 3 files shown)


nuwang:
This is the commit in which the changes occurred: https://github.com/galaxyproject/galaxy/pull/16795
I think your fix for this is correct. The whole `<uuid>.ep.interactivetool` part still seems to be a bit redundant to me, why not just `<uuid>.interactivetool` and avoid the extra dns lookup for the `ep` part? But that seems like an issue for another day.

nuwang (approved):

PR #1774, opened 2024-02-09
Update local_tool_conf.xml

cat-bro:
@mthang does this look right? Is it a test tool or open to all users?


mthang:
@cat-bro That's correct. I am the only one who will work / investigate on this tool. Thanks

PR #1785, opened 2024-02-21
Configure ulimits for max open files

nuwang:


cat-bro (approved):
Looks good! Thanks @nuwang

PR #1789, opened 2024-02-22
gxadmin local query jobs by object store id

cat-bro:
Add a query to see latest jobs with data from various object store backend IDs.

Also remove local_query-all-jobs from this file because this has been replaced by 'gxadmin query jobs'


nuwang (approved):
lgtm!

PR #1793, opened 2024-02-22
bwa_mem cannot run on pulsar-QLD

cat-bro:


cat-bro:
This is incorrect anyway. The index not found came from cvmfs.

PR #1801, opened 2024-02-26
Increase pulsar preprocess_action_max_retries from 20 to 30

cat-bro:
To address jobs with large input failing because the data does is not transferred in time


cat-bro:
@nuwang this is also for the benefit of large files being transported with curl

nuwang (approved):
lgtm!

PR #1825, opened 2024-03-07
Reduce phm1 realmemory to available memory

jlqfab:
After rebuild of pulsar-high-mem1 it has less free memory than what is specified by slurm's `RealMemory` which prevents slurm from utilising the node. Reducing `RealMemory` to match what's available.


jlqfab:
accidently mixed with unmerged PR

PR #1832, opened 2024-03-10
Add optional upstream proxy for memray

nuwang:


templates/nginx/galaxy.j2
```diff
@@ -1,9 +1,9 @@
-{% if num_gunicorn_processes == 2 %}
 upstream galaxy {
-    server unix://{{ galaxy_mutable_config_dir }}/gunicorn1.sock;
-    server unix://{{ galaxy_mutable_config_dir }}/gunicorn2.sock;
-}
+{% for i in range(1, num_gunicorn_processes+1) %}
+    server unix://{{ galaxy_mutable_config_dir }}/gunicorn{% if num_gunicorn_processes > 1 %}{{i}}{% endif %}.sock weight={{100-debug.memray.weight/num_gunicorn_processes}};
```
cat-bro, line 3: do you think jinja will do the maths here or make it a literal string? `{{100-debug.memray.weight/num_gunicorn_processes}}`

cat-bro (approved):

templates/nginx/galaxy.j2
nuwang: Should work. Anything within brackets is just python, so it should evaluate it as a python expression. Just in case, I tested it here: http://jinja.quantprogramming.com/

PR #1838, opened 2024-03-14
Change mount settings

cat-bro:
The reason we started setting actimeo at ETCA was that files created and shown as available in the history panel would not become available until 10-20 seconds later. From memory, using actimeo=2 for the user volumes was adequate. I don't know if we have a use case for setting actimeo=0 on /mnt/tmp - switching to default is worth a try to see if it will improve performance.

### DIFF
--- a/group_vars/galaxy_etca.yml
+++ b/group_vars/galaxy_etca.yml
@@ -47,22 +47,22 @@ galaxy_server_and_worker_shared_mounts: # Everything mounted on galaxy, galaxy_h
     src: "{{ hostvars['galaxy-job-nfs']['internal_ip'] }}:/mnt/tmp"
     fstype: nfs
     state: mounted
-    opts: 'noatime,actimeo=0,defaults'
+    opts: 'noatime,defaults'
   # galaxy-user-nfs
   - path: /mnt/user-data-volA # 150T volume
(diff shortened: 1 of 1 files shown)


jlqfab (approved):

nuwang:
Since /mnt/tmp doesn't need to be accessible from a shared location, agree that actimeo is likely to be more harmful than helpful. Why not remove specifying it altogether?

cat-bro:
/mnt/tmp needs to be accessible from all of the workers, galaxy and galaxy-handlers. I've removed actimeo=0 from the mount options for /mnt/tmp. It does still need to be on any user-data volume with an object store backend that has weight>0 or has recently had weight>0, but I think it doesn't need to be 0.

nuwang:
Each file in /mnt/tmp is written to/read by one worker at a time right? The same file does not need be accessible from multiple workers at the same time? We need actimeo when multiple workers need to see the same file?

cat-bro:
The biggest use case for has been galaxy-handlers (or the workers) deliver a file to user-data, tell galaxy it's ready, galaxy says it's ready (green dataset with an eyeball icon) but when someone clicks on the eyeball galaxy VM itself thinks the file is not there, hence setting it on galaxy + handlers and workers for user volumes

PR #1847, opened 2024-03-19
Less nfs caching of user data mounts

cat-bro:
Only the galaxy VM needs actimeo set for the user data volumes. It doesn't make sense that any of the other VMs would need this.

These volumes are mounted on 12 clients. Is there any difference between all 12 of them having actimeo set, or only one of them? If not, this change is not helpful and this PR can be closed.

### DIFF
--- a/group_vars/galaxy_etca.yml
+++ b/group_vars/galaxy_etca.yml
@@ -31,6 +31,8 @@ pawsey_file_mounts_available: True # assume true unless set to false
 qld_file_mounts_path: /mnt/user-data-qld
 pawsey_file_mounts_path: /mnt/user-data-pawsey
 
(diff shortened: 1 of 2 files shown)


cat-bro:
@nuwang @jlqfab as described above - I'm not sure if it an effective change to switch off the actimeo settings on 11 out of 12 clients for the user disks. Switching it off on 12 out of 12 is not an option.

nuwang (commented):
This is a sensible change. Should be activated for web + handlers I think. e.g. worker finishes job, but handler may not see files immediately.

cat-bro:
This will work on galaxy and galaxy-handlers VMs due to galaxy-handlers inheriting the value from host_vars/galaxy.usegalaxy.org.au

nuwang (approved):
Looks good!

PR #1854, opened 2024-03-21
vary nginx logrotate file to keep logs forever (dev)

cat-bro:
This is a setup for dev to see if it has the desired result.

Two weeks of nginx logs is 1.1GB in size so keeping them forever on the root disk is not an option.

There is a lot of space on /mnt on the galaxy VM so we could probably keep them for several years or just this year to get access data. This is a few lineinfiles on the nginx logrotate conf that adds a date to the files, stores them in an olddir which has to be on the same device (apparently logrotate will protest if it is not) and copies the files to a directory on /mnt where they will not be rotated out but accumulate instead.

If this is approved and works on dev I'll want to put it on production (and remove it from dev where space on /mnt is not abundant).

before:
```
/var/log/nginx/*.log {
	daily
	missingok
	rotate 14
	compress
	delaycompress
	notifempty
	create 0640 www-data adm
	sharedscripts
	prerotate
		if [ -d /etc/logrotate.d/httpd-prerotate ]; then \
			run-parts /etc/logrotate.d/httpd-prerotate; \
		fi \
	endscript
	postrotate
		invoke-rc.d nginx rotate >/dev/null 2>&1
	endscript
}
```

after:
```
/var/log/nginx/*.log {
	olddir /var/log/nginx/olddir
	dateext
	daily
	missingok
	rotate 14
	compress
	delaycompress
	notifempty
	create 0640 www-data adm
	sharedscripts
	prerotate
		if [ -d /etc/logrotate.d/httpd-prerotate ]; then \
			run-parts /etc/logrotate.d/httpd-prerotate; \
		fi \
	endscript
	postrotate
		cp /var/log/nginx/olddir/* /mnt/var/log/nginx/
		invoke-rc.d nginx rotate >/dev/null 2>&1
	endscript
}
```

### DIFF
--- a/dev_playbook.yml
+++ b/dev_playbook.yml
@@ -96,3 +96,27 @@
       become: yes
       become_user: root
 
+    # Nginx logrotate: keep nginx logs forever on /mnt disk
+    - name: Make sure extra nginx log dirs exist
+      file:
+        state: directory
+        path: "{{ item }}"
+      with_items:
+        - nginx_log_olddir_root
+        - nginx_long_term_log_dir
+    - name: Add lines to main block in nginx logrotate conf
+      lineinfile:
+        path: /etc/logrotate.d/nginx
+        line: "{{ item }}"
+        insertafter: ".*{"
+      with_items:
+        - "\tdateext"
+        - "\tolddir {{ nginx_log_olddir_root }}"
+    - name: Add lines to postrotate task in nginx logrotate conf
+      lineinfile:
+        path: /etc/logrotate.d/nginx
+        line: "{{ item }}"
+        insertafter: "\tpostrotate"
+      with_items:
+        - "\t\tcp {{ nginx_log_olddir_root }}/* {{ nginx_long_term_log_dir }}/"
+
--- a/dev_playbook.yml
+++ b/dev_playbook.yml
@@ -102,8 +102,8 @@
         state: directory
         path: "{{ item }}"
       with_items:
-        - nginx_log_olddir_root
-        - nginx_long_term_log_dir
+        - "{{ nginx_log_olddir_root }}"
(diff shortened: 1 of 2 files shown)


dev_playbook.yml
```diff
@@ -96,3 +96,27 @@
...
       become_user: root
 
+    # Nginx logrotate: keep nginx logs forever on /mnt disk
+    - name: Make sure extra nginx log dirs exist
+      file:
+        state: directory
+        path: "{{ item }}"
+      with_items:
+        - nginx_log_olddir_root
+        - nginx_long_term_log_dir
```
nuwang, line 106:
```suggestion
        - "{{ nginx_log_olddir_root }}"
        - "{{ nginx_long_term_log_dir }}"
```

dev_playbook.yml
```diff
@@ -96,3 +96,27 @@
...
+    - name: Add lines to main block in nginx logrotate conf
+      lineinfile:
+        path: /etc/logrotate.d/nginx
+        line: "{{ item }}"
+        insertafter: ".*{"
+      with_items:
+        - "\tdateext"
```
nuwang, line 113: Not sure what dataext does but I assume that's necessary.

dev_playbook.yml
cat-bro: Thanks!

dev_playbook.yml
cat-bro:
```
dateext
Archive old versions of log files adding a daily extension like YYYYMMDD instead of simply adding a number. The extension may be configured using the dateformat option.
```

At the moment our log files are numbered so without adding the date we'd have no way to ID them long term, plus they'd overwrite each other

nuwang (approved):
lgtm!

PR #1856, opened 2024-03-21
clean_tmpdisk.py: filter squeue for worker_node

cat-bro:


nuwang (commented):
I think the `node != worker_node` line below will also not be needed since it'll always be the selected node.

cat-bro:
I agree, but I like keeping it and having the double check

nuwang (approved):

PR #1858, opened 2024-03-22
Copy access/error logs to /mnt in nginx postrotate script

cat-bro:
This works on dev

### DIFF
--- a/galaxy_playbook.yml
+++ b/galaxy_playbook.yml
@@ -119,4 +119,4 @@
         line: "{{ item }}"
         insertafter: "\tpostrotate"
       with_items:
-        - "\t\tcp {{ nginx_log_olddir_root }}/* {{ nginx_long_term_log_dir }}/"
+        - "\t\tcp -up {{ nginx_log_olddir_root }}/* {{ nginx_long_term_log_dir }}/"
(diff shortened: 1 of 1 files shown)


galaxy_playbook.yml
```diff
@@ -96,3 +96,27 @@
...
+      with_items:
+        - "\tdateext"
+        - "\tolddir {{ nginx_log_olddir_root }}"
+    - name: Add lines to postrotate task in nginx logrotate conf
+      lineinfile:
+        path: /etc/logrotate.d/nginx
+        line: "{{ item }}"
+        insertafter: "\tpostrotate"
+      with_items:
+        - "\t\tcp {{ nginx_log_olddir_root }}/* {{ nginx_long_term_log_dir }}/"
```
jlqfab, line 122: `cp {{ nginx_log_olddir_root }}/* {{ nginx_long_term_log_dir }}/` is copying everything from the directory and will overwrite existing copies, thus creating unnecessary overhead. It would be better to use `cp -up` to only copy and overwrite files that have been updated (`-u`) and preserve timestamps, ownership and mode (`-p`).

jlqfab (approved):

PR #1861, opened 2024-03-22
Set max_map_count to 1,000,000 for all qld high memory pulsars

cat-bro:


cat-bro:
@jlqfab what do you think? We did this for pqhm1 only and we were using this for other tools too.. does it stand to reason that it is OK to run on any of the large pulsars?

jlqfab:
This will allow processes to increase their memory footprint, it can potentially increase memory consumption by an application which could reduce server performance overall. If it works ok on pqhm1 being shared with other tools, theoretically it should be ok on all the large memory pulsars, just be aware that it could allow a single application that dynamically allocates memory to hog the memory and limit availability of memory to the OS. In other words, you might want to be able to choose what you let run in that environment, which may be a case for limiting this setting to a few pulsars, or being less generous with the `max_map_count` value on some. Or try it and if it impacts performance, limit it.

cat-bro:
Thanks JL. For the interests of scheduling I'm going to go ahead with this change, as limiting maxquant to one VM will be bad for the queues. I haven't seen evidence of more bad jobs on pqhm1 than the others, so fingers crossed.

Issue #1849, opened 2024-03-20
Slurm nodes that go into drain 'DUE TO JOB NOT ENDING WITH SIGNALS'


cat-bro:
This is the mode of most slurm nodes that go into drain state. This happens mostly with the local slurm workers and to an extent with the qld high memory slurm nodes.

I usually check sacct to get the name of the job (which contains the galaxy ID), then `ps aux | grep <galaxy_job_id>` which would show processes if the galaxy job were still running. I have never found the the galaxy job is still running.

(1) Are there slurm settings that we could change that would allow the job more time to stop before slurm declares that is is not stopping and puts the node in drain state?

(2) Failing this, could we automate the resumption of slurm nodes that are draining under this circumstance?

cat-bro:
Since setting UnkillableStepTimeout to 180 this has been happening far less frequently

PR #1863, opened 2024-03-22
set unkillablesteptimeout in slurm conf

cat-bro:
For issue https://github.com/usegalaxy-au/infrastructure/issues/1849

Slurm nodes go into drain state due to jobs not ending with signals. The UnkillableStepTimeout defaults to 60 seconds. Hopefully raising it to 180 will stop the slurm nodes going into drain state as frequently.


cat-bro:
I've run the dev_update_slurm playbook using this branch and can report that it does not kill slurm

jlqfab (approved):

PR #1866, opened 2024-03-25
raise mem for DATA_FETCH to 12

cat-bro:
This will increase queue times for data fetch when it's busy, on the other hand no upload job should fail so the change is necessary.


nuwang:
Which type of uploads use this much memory?

cat-bro:
Autodecting the datatype when it will be 'qza' type. There might be other types that are difficult for sniffers as well. I feel that 12GB is too much to allocate for data fetch. I'd like to be able to allocate 4.

nuwang:
That sounds like a bug. Autodetecting a datatype should be fairly efficient and quick. So it's most likely loading 12GB into memory?

cat-bro:
Does this link work? https://drive.google.com/file/d/1CvmZ1ONAi1dl2_TnzVm4Hc4gG6eSz7a5/edit

I'm not sure if it will still fail with this file because we have raised the memory limit

PR #1883, opened 2024-04-04
pin pycurl for pulsar

cat-bro:
pulsar-qld-gpu1 had the same problem with cert location as pulsar-high-mem1. I've downgraded pycurl to 7.45.2 and now it's OK. For now this will need to be pinned.

There is a bug report but I didn't find it last week - only saw it just now by specifically searching for pycurl 7.45.3 errors https://github.com/pycurl/pycurl/issues/834


jlqfab (approved):
Nice find!

PR #1907, opened 2024-04-11
Additions to configure pulsar-qld-gpu2 as a pulsar

jlqfab:


cat-bro:
6 near-identical host files are going to be unwieldy and hard to maintain. It would be better to have a group file for most of these variables. The slurm bit could be:

```
slurm_nodes:
    - name: "{{ ansible_hostname }}"
      NodeAddr: "{{ hostvars[ansible_hostname]['ansible_ssh_host'] }}"
      CPUs: 64
      RealMemory: 596400
      State: UNKNOWN
```

et cetera. There is very little that could not be shared.

cat-bro (dismissed):

cat-bro:
This is good to go but before we run all 6 we could make a group_vars/pulsar_qld_gpu.yml for common vars

jlqfab:
Agreed. I was thinking that there's too much duplication.

PR #1923, opened 2024-04-22
add headers to /_x_accel_redirect/ block in nginx template

cat-bro:
EU and main both have this. I'm wondering if a recent problem with content-length not being defined things sent to pulsar might have to do with the absence of this in our own config.


nuwang:
Not sure whether these particular settings will have an effect, since they affect CORS and content type sniffing, but no harm in adding them I guess.

nuwang (approved):

PR #1934, opened 2024-04-26
rebuild pulsar-QLD

cat-bro:
rebuild pulsar-QLD with a 10TB volume for pulsar + job directories and a 3TB volume for custom indices. Remove pulsar-QLD-queue but keep the NFS VM for custom-indices. Rebuild the VMs with ubuntu 22.04


cat-bro:
This needs to be merged to be tested

PR #1957, opened 2024-05-08
make max length and db path context variables for alphafold so that max length can be extended for group members

cat-bro:
Hi @neoformit, this should work to allow members of an extra group to run alphafold with bigger inputs. They would have to be members of the base 'Alphafold' group as well as 'Alphafold_plus' (or whatever name) and the new group would need to have an associated role. Pick whatever number is suitable for the aa length max in the new group.


neoformit:
This looks great, thanks Catherine. I just noticed that max seq count is hard-coded to 10 so will need to update that and release a new tool version to the TS before this can be done.

neoformit:
@cat-bro If this looks good to you, I think if you merge this and install new AF2 revision on GA, we are good to go :slightly_smiling_face:

neoformit (approved):

neoformit:
@cat-bro are we good to merge and install?

PR #1965, opened 2024-05-13
Fail bwa_mem2 jobs submitted with more than 150G input

cat-bro:


jlqfab (approved):
Good idea to tell them to compress their inputs.

PR #1971, opened 2024-05-23
add uwe to backup and refactor machine users

cat-bro:
Add Uwe as a non-sudo user on galaxy-backup

Refactor machine users so that users in `host_machine_users` will be added to the list or will supersede entries with the same name in all.yml.


jlqfab (approved):
Looks good!

PR #1991, opened 2024-06-05
Add tpv availablity score ranking but do not enable it yet

cat-bro:
Add some more log statements to ranking function: what is being ranked and what the final ordered destinations are. Log space on galaxy-handlers has not been an issue.

Add function to replace destination_usage_proportion. destination_availability_score(destination) will return the maximum number of jobs of the type that could be scheduled given available resources at the destination. It is an attempt to improve the scheduling of jobs that have a high memory requirement compared to number of CPUs. Up to this point ranking has always considered only CPUs.  This is commented out so that I can collect some logs from the previous method before switching them.

### DIFF
--- a/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/default_tool.yml.j2
+++ b/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/default_tool.yml.j2
@@ -32,6 +32,7 @@ tools:
         log.info("++ ---tpv rank debug: 1 or fewer destinations: returning candidate_destinations")
         final_destinations = candidate_destinations
       else:
+        log.info(f"++ ---tpv rank debug: ranking destinations for job with tool {job.tool_id}, cores: {cores}, mem: {mem}")
         import random
         from sqlalchemy import text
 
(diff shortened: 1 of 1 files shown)


cat-bro:
tested on dev

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/default_tool.yml.j2
```diff
@@ -65,15 +66,33 @@ tools:
 
           def destination_usage_proportion(destination):
             if not destination.context.get('destination_total_mem') or not destination.context.get('destination_total_cores'):
-                    raise Exception(f"++ ---tpv rank debug: At least one of destination_total_mem, destination_total_cores is unavailable")
+              raise Exception(f"++ ---tpv rank debug: At least one of destination_total_mem, destination_total_cores is unavailable")
             destination_total_cores = destination.context.get('destination_total_cores')
             return_value = sum([db_queue_info.get(destination.id, {}).get(state, {}).get('sum_cores', 0) for state in ['queued', 'running']])/destination_total_cores
```
nuwang, line 71: Since state 'queued' and 'running' are already filtered out at the sql level, and both states are always combined into a single value, is there an advantage in breaking things down by queued and running in the db_queue_info dict?

cat-bro: No, that is a good point

cat-bro: part of the reason for this is that I would like to write a generic helper function for tpv around the parameters 'tpv_cores', 'tpv_mem', 'tpv_gpus'. I am not considering a difference between queued and running but someone else might, and the return value would include that information

```
def get_queue_info(app):
    from sqlalchemy import func, text, Float
    from galaxy.model import Job
    query = app.model.context.query(
        Job.destination_id,
        Job.state,
        func.count(Job.id).label('job_count'),
        func.sum(
            func.cast(text("encode(destination_params, 'escape')::json ->> 'tpv_cores'"), Float)
        ).label('total_cores'),
        func.sum(
            func.cast(text("encode(destination_params, 'escape')::json ->> 'tpv_mem'"), Float)
        ).label('total_mem'),
        func.sum(
            func.cast(text("encode(destination_params, 'escape')::json ->> 'tpv_gpus'"), Float)
        ).label('total_gpus'),
    ).filter(
        Job.state.in_(['queued', 'running'])
    ).group_by(
        Job.destination_id,
        Job.state
    )
    results = query.all()
    < construct db_queue_info from results and return db_queue_info >
```

nuwang: Right, that makes sense. Let's leave it as is then. We could also consider making the job states a parameter to get_queue_info.

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/default_tool.yml.j2
```diff
@@ -65,15 +66,33 @@ tools:
...
             log.info(f"++ ---tpv rank debug: returning usage proportion value for destination {destination.id}: {str(return_value)}")
             return return_value
 
+          def destination_availability_score(destination):
+              if not destination.context.get('destination_total_mem') or not destination.context.get('destination_total_cores'):
+                raise Exception(f"++ ---tpv rank debug: At least one of destination_total_mem, destination_total_cores is unavailable")
+              destination_cores_committed = sum([db_queue_info.get(destination.id, {}).get(state, {}).get('sum_cores', 0) for state in ['queued', 'running']])
+              destination_mem_committed = sum([db_queue_info.get(destination.id, {}).get(state, {}).get('sum_mem', 0) for state in ['queued', 'running']])
+              available_cores = destination.context['destination_total_cores'] - destination_cores_committed
+              available_mem = float(destination.context['destination_total_mem']) - (float(destination_mem_committed)/1024)
```
jlqfab, line 81: So, `destination_mem_committed` and `destination.context['destination_total_mem']` are using a different base (i.e. KB vs GB)?

cat-bro: because the query is taking the value from slurm native specification, which is in MB

jlqfab (approved):

Issue #1942, opened 2024-04-30
Add training history tag


cat-bro:
Schedule any jobs from a history with a tag named 'training' to the training resources.

This would use the same logic as the 'singularity' tag detection rules for test users in users.yml.

It would need to be on the default user. It might make sense to move all of the training role logic from roles.yml to users.yml

PR #1993, opened 2024-06-06
remove roles to merge training role rules with training tag rules

cat-bro:
For consideration: The aim is that somebody could add the tag 'training' to their history and get the same treatment as somebody in a training group, which could be useful for testing a tutorial ahead of time.

For issue #1942


files/galaxy/dynamic_job_rules/production/total_perspective_vortex/users.yml
```diff
@@ -1,4 +1,82 @@
...
+                training_rules_apply = True
+            except Exception as e:
+              log.debug('--tpv log debug: Exception checking training status of job')
+              log.debug(str(e))
+              
+            if training_rules_apply:
+              include_tags = ['pulsar']
+              exclude_tags = ['pulsar-training-large', 'training-exempt']
+              slurm_preferred_for_training = pulsar_score is not None and pulsar_score < pulsar_score_slurm_threshold
+              training_job_matches_specs = (not slurm_preferred_for_training) and helpers.tag_values_match(entity, include_tags, exclude_tags)
```
nuwang, line 48: Since only this line is different between slurm and pulsar, maybe it would make sense to move this code to a seperate file that we just import. A possible future enhancement. I imagine more scenarios like this will pile up.

nuwang (dismissed):
This looks good to me.

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/users.yml
cat-bro: I agree - most of the code in the 3 blocks is identical.

PR #2009, opened 2024-06-12
Update local_tool_conf.xml

mthang:
Remove test label for all fgenesh tools from the local_tool_conf file


cat-bro:
Without the test label the visibility filtering would disappear, so this cannot be merged until we update visibility filtering rules

mthang:
Do I need to do anything on the visibility filtering rules? should I close this PR and wait until the visibility filtering rules get updated?

PR #2017, opened 2024-06-18
[WIP] Add gpu specific requirements to pulsar qld gpu playbook

jlqfab:
Initial untested commits for configuring Nvidia GPU with docker support for Nectar pulsar-qld-gpu's.

### DIFF
--- a/one-offs/initial_nvidia_config_and_disable_cgroups_v2_playbook.yml
+++ b/one-offs/initial_nvidia_config_and_disable_cgroups_v2_playbook.yml
@@ -1,0 +1,30 @@
+- hosts: all
+  become: true
+  tasks:
+  - name: Install nvidia kernel modules, drivers, utils (nvidia-smi) and cuda-toolkit
+    apt:
+      pkg:
+        - "linux-modules-nvidia-{{ nvidia_driver_version }}-server-generic"
+        - "nvidia-driver-{{ nvidia_driver_version }}-server"
+        - "nvidia-utils-{{ nvidia_driver_version }}-server"
+        - "nvidia-cuda-toolkit"
+      state: present
+  - name: Reboot for kernel updates and kernel version to match nvidia modules - nvidia kernel modules will be loaded on reboot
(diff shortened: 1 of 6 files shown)


one-offs/initial_nvidia_config_and_disable_cgroups_v2_playbook.yml
```diff
@@ -0,0 +1,30 @@
+- hosts: all
```
cat-bro, line 1: hosts: all?

roles/nvidia-ctk/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+
+- name: Install nvidia-container-toolkit
+  apt:
+    name: nvidia-container-toolkit
+    state: present
+
+- name: Configure the container runtime and restart Docker
+  shell:
+    nvidia-ctk runtime configure --runtime=docker
+  notify: restart docker
```
cat-bro, line 46: is there a handler somewhere that will handle this?

one-offs/initial_nvidia_config_and_disable_cgroups_v2_playbook.yml
jlqfab: intention is to run with a `--limit` for testing, but should be defined properly, as you say.

roles/nvidia-ctk/tasks/main.yml
jlqfab: There will be. Thanks - forgot to add it.

jlqfab:
@cat-bro the `nvidia-ctk` role in `pulsar-qld-gpu_playbook.yml` requires both Nvidia and Docker to already be installed. Nvidia configuration for MIG mode and reboots for kernel modules only need to be done once, but this needs to be done prior to the main playbook being run. Hence, addition of a check for nvidia at the start of the main playbook.

cat-bro (approved):

PR #2020, opened 2024-06-19
fixes to pulsar-qld-gpu playbook and nvidia-ctk role

jlqfab:
`nvidia-detector` only tested if nvidia GPU present, changed to using `nvidia-smi `which will fail of the one-offs hasn't been run.
`nvidia-ctk` role fixed - need to use a .asc extension rather than .gpg for repo key, this means that don't need to add an additional gpg dearmor step.


cat-bro (approved):
Looks good TM!

PR #2039, opened 2024-06-26
Revert "Update anndata import rule"

cat-bro:
Reverts usegalaxy-au/infrastructure#2038


cat-bro:
it's equivalent, no need to revert

PR #2044, opened 2024-06-29
Add walle to requirements + dev

cat-bro:
This is a draft. All that it does so far is add the [walle ansible role](https://github.com/usegalaxy-eu/WallE) to requirements.yml.  This role, developed by Mira of Galaxy Europe, will help defend interactive tools from misuse

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -80,3 +80,5 @@ roles:
 - name: geerlingguy.redis
   version: 1.8.0
 
+- name: usegalaxy_eu.walle  # TODO: ALPHEBATISE THIS FILE IT IS HARD TO READ
+  version: 0.0.2


cat-bro:
probably doesn't matter whether this goes onto gunicorn machine or handlers machine - needs access to jwds, config files and db which are accessible from both. handlers might make more sense

cat-bro:
using walle requires that we copy this script https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/roles/usegalaxy-eu.bashrc/files/galaxy_jwd.py, probably just wget, into the same directory that walle uses, so it would make sense for this directory to be within /mnt/galaxy or something that both VMs have access to. This script can be used for a lot of "get jwd path", "get dataset path" type bash stuff as well

PR #2055, opened 2024-07-10
configure pulsar-qld-gpu5 host_vars job_conf_limits and TPV destination

jlqfab:


cat-bro (approved):

jlqfab:
Is it necessary to set it offline in TPV before it is built, given it is only being used for testing at the moment?

PR #2058, opened 2024-07-11
create playbook for pulsar-qld-gpu-dev

jlqfab:


cat-bro:
Does this need a separate playbook?

jlqfab:
I didn't think so originally, but saw other dev playbooks and noted they included `group_vars/dev.yml` and don't call the `slg.galaxy_stats` role. So thought it might be needed. Which probably also means I should change the DNS domain of pulsar-qld-gpu-dev to be `.gvl.org.au`

jlqfab:
As per @cat-bro's comment. This playbook is not required - can directly use the `pulsar-qld-gpu` playbook.

PR #2059, opened 2024-07-11
[WIP] Remove subdomains role

cat-bro:
tedious busy work. not ready for merge but will need running on dev from this branch to check whether it actually works.


cat-bro:
Having gone through the docs and eu's configuration to try to make this work, it's really difficult to understand why the subdomains role was not kept and refactored that instead of requiring the role users to refactor everything

PR #2062, opened 2024-07-12
Update tools.yml

igormakunin2:
Increase slot and memory allocation for funannotate_predict


cat-bro (approved):
Thanks @igormakunin2 !

PR #2066, opened 2024-07-15
Try sending helixer to pulsar-qld-gpu

TomHarrop:


TomHarrop:
helixer still has a `test` label so this shouldn't affect users.

cat-bro (approved):

PR #2079, opened 2024-07-17
exclude pulsar-qld-gpu-dev from pulsar-qld-gpu playbook

jlqfab:
exclude pulsar-qld-gpu-dev from pulsar-qld-gpu_playbook as it is a member of pulsar_qld_gpus but has its own playbook to give devs sudo


cat-bro (approved):
nice! new lingo

PR #2085, opened 2024-07-22
galaxy singularity common cache

cat-bro:
To fix the issue when the first run of a tool fails after the container has been auto fetched. SINGULARITY_CACHEDIR needs to be off the root disk and the path needs to exist on the workers so it may as well be common. On production galaxy it will be /mnt/tools/singularity_data.

This change will improve the tool installation process and make it easier to mark existing tools as singularity enabled.

[EDIT] I mistakenly wrote /mnt/galaxy/singularity_data


jlqfab:
From what I can see, the galaxy production head node defines `galaxy_tools_indices_dir` as `/mnt/tools`, as opposed to `galaxy_root`  on dev and staging. This would result in the head node having the singularity cache in `/mnt/tools/singularity_data` as opposed to `/mnt/galaxy/singularity_data`.

cat-bro:
yep, that's what I meant. Typo (now corrected)

jlqfab (approved):

PR #2088, opened 2024-07-24
use production alphafold DBs on pulsar-qld-gpus

jlqfab:
@neoformit has consolidated the AF2 DBs to a single new volume. These changes replace the existing Alphafold Database mounts with a single 'production' alphafold database mount.

### DIFF
--- a/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/dev_destinations.yml.j2
+++ b/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/dev_destinations.yml.j2
@@ -192,7 +192,7 @@ destinations:
     rules:
       - id: pulsar_destination_docker_rule
         params:
-            docker_volumes: $job_directory:ro,$tool_directory:ro,$job_directory/outputs:rw,$working_directory:rw,/data/alphafold_databases:/data:ro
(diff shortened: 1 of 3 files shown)


jlqfab:
Note that these changes should only affect `alphafold_test` jobs.

Additional work needs to be done to the TPV to ensure that production alphafold jobs can be executed on azure and pulsar-qld-gpus. Specifically regarding `toolshed.g2.bx.psu.edu/repos/galaxy-australia/alphafold2/alphafold/.*:` where `docker_volumes` mounts `/data/alphafold_databases` (azure), while pulsar-qld-gpus use the mount `/data/alphafold`.

Options:
- specify path to alphafold databases in TPV destinations;
- make all mount points the same `/data/alphafold_database`

neoformit:
I think `docker_volumes` is already overwriting this in the destination, that's how it currently works and why I suggested getting rid of the `docker_volumes` at the tool level as it's a bit confusing/misleading?

neoformit:
It would also make sense to make all mount points the same, but I guess this point is redundant if we drop Azure for Nectar soon anyway.

neoformit (approved):
I think the mount points will be fixed at the destination level anyway. Worst case scenario we break test alphafold and make another PR

PR #2097, opened 2024-07-30
Toolbox filter on AAGI tools

neoformit:
Restrict tools with the `aagi` label to users with the `aagi` role
(QCIF project)


neoformit:
Thanks @cat-bro I think it's good to go now

cat-bro (approved):

PR #2130, opened 2024-08-28
Remove task for gxctl bash function

cat-bro:
Since some versions ago we have been using 'sudo galaxyctl' instead of this


jlqfab (approved):
bye bye gxctl

PR #2140, opened 2024-09-03
Update galaxy and staging object store config files to yaml format

cat-bro:


jlqfab (dismissed):
Looks good! Staging's object conf has weird naming where `staging_objects_1` points to `/mnt/galaxy/data-2` and `staging_objects_2` points to `/mnt/galaxy/data`, but it matches the xml config so that's ok.

PR #2150, opened 2024-09-08
add aws gateway to etca mounts

cat-bro:


group_vars/galaxy_etca.yml
```diff
@@ -104,6 +104,11 @@ galaxy_server_and_worker_shared_mounts: # Everything mounted on galaxy, galaxy_h
     fstype: nfs
     opts: 'noatime,defaults'
     state: "{{ 'mounted' if qld_file_mounts_available else 'absent_from_fstab' }}"
+  - path: /mnt/aws_gateway
+    src: 192.168.205.17:/galaxy-aws-gw-test-bucket-2024-09-06-2fjsq
+    opts: 'nolock,hard,noatime,defaults'
```
jlqfab, line 109: Should we include `defaults` option here? If we need it, then it should be in the order `defaults,nolock,hard,noatime` as the order matters - latter options override earlier options.
Also note that on galaxy-w7 we mounted with options `nolock,hard,noatime` without `default`.

cat-bro: Good point about order.

Also I'm wondering if 'defaults' is redundant: surely 'defaults,A,B=C' is the same as 'A,B=C'. I'll take this out.

jlqfab (approved):
LGTM

PR #2158, opened 2024-09-11
Adding builtins for collections based on main

TomHarrop:
These are all built ins that are installed on usegalaxy.org, but I haven't tested any of them.


cat-bro:
GA has a pick_value already installed from the toolshed. GA should only have one of these

cat-bro (approved):

PR #2163, opened 2024-09-12
Dev handlers to run nginx upload module

cat-bro:
@nuwang @jlqfab the main change is the reduced nginx conf for a handlers VM which can be reused for production.


jlqfab (dismissed):

jlqfab (approved):
ah ok. so this is the reduced nginx config.

host_vars/dev-handlers.gvl.org.au.yml
```diff
@@ -26,17 +26,26 @@ galaxy_fetch_dependencies: false
...
   galaxy_user: "{{ galaxy_user.name }}"
   app_server: gunicorn
   virtualenv: "{{ galaxy_venv_dir }}"
-  gunicorn:
-    enable: false
   celery:
     concurrency: 2
     loglevel: DEBUG
+  gunicorn:
+    - bind: "unix:{{ galaxy_mutable_config_dir }}/gunicorn.sock"
```
nuwang, line 42: The gunicorn sock might be in a shared folder, so this may need to be renamed.

cat-bro: Thanks, good catch!

jlqfab (approved):

PR #2171, opened 2024-09-13
switch on gunicorn and nginx on galaxy-handlers

cat-bro:
Add gunicorn process to galaxy-handlers and switch one of the pulsars to get files from galaxy-handlers instead of galaxy


jlqfab (approved):
Looks ok to me :-)

PR #2173, opened 2024-09-16
all pulsar runners use galaxy-handlers

cat-bro:


cat-bro:
In the future this will probably either be shared between the two VMs (some pulsars using each one) or delegated to anther VM.

jlqfab (approved):
Nice!

PR #2175, opened 2024-09-16
add geerlingguy.jenkins role for jenkins-2024 playbook

jlqfab:


jlqfab:
also add geerlingguy.jenkins role to ansible requirements

PR #2179, opened 2024-09-17
Add michaeld keys

madsi1m:
also move filesender app to "Send Data" on dev


group_vars/dev.yml
```diff
@@ -39,3 +39,10 @@ host_machine_users:
         - galaxy_admin
       key: files/keys/tom.pub
       email: "{{ tom_email }}"
+    - name: michaeld #aarnet
```
cat-bro, line 42: no need to change this file unless overriding what is in all.yml

files/keys/michaeld.pub
```diff
@@ -0,0 +1 @@
+ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINzXuaBkXivTxDMpuF9z0voWM6mBRr5z0Laf0Y92cruG michaeld@galaxy
```
cat-bro, line 1: If I remember right this is too modern for the VMs that are pre-22.04, dev for example. Do you have and rsa key? https://www.ssh.com/academy/ssh/keygen

cat-bro (approved):

PR #2181, opened 2024-09-17
test ubuntu version for pre-22.04 in common role task

jlqfab:


jlqfab:
Adding confitional to task as this role doesn't work on Ubuntu 24.04, and according to the comment in the role is only required for Ubuntu versions prio to 22.04.

cat-bro (approved):

PR #2183, opened 2024-09-17
common packages deal with ubuntu 24.04 differences

jlqfab:
the Ubuntu package `libgl1-mesa-glx` is obsolete since 23.10. Replace this on Ubuntu 24.04 with `libgl1` and `libglx-mesa0` packages which provide equivalent behaviour.


cat-bro (approved):
nice! good diplomacy

PR #2187, opened 2024-09-18
Add michaeld_email to vault for machine_users

jlqfab:
We need to add the email to the vault for machine_users


cat-bro (approved):
Thanks @jl!

PR #2188, opened 2024-09-18
ignore all geerlingguy roles - installed via requirements.yml

jlqfab:
Rather than ignoring the installed geerlingguy roles on a one by one case, ignore all of them by default.


cat-bro (approved):
ok I won't write a local tribute role called geerlingguy-is-rad

PR #2190, opened 2024-09-19
Diffbind and mirdeep tpv

cat-bro:
File size and singularity rules for diffbind, more mem for mirdeep2_mapper which is fine on pulsar. This also adds a bunch of rule IDs where they were missing


jlqfab (approved):
LGTM

PR #2193, opened 2024-09-22
Add minio and aws_gateway object store entries

cat-bro:
See [sample object store conf file](https://github.com/galaxyproject/galaxy/blob/release_24.1/lib/galaxy/config/sample/object_store_conf.sample.yml)


cat-bro:
I imagine there might be some iterating to get optimal configurations for minio_test especially.

It probably won't break galaxy but I'll test restarting an individual job handler and revert this if the handler won't start.

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/users.yml
```diff
@@ -115,6 +115,17 @@ users:
...
           - slurm
+      rules:
+        - id: star_minio_test_rule
+          if: |
+            'fastqc' in tool.id
+          params:
+            object_store_id: minio_test
+        - id: star_aws_gateway_test_rule
+          if: |
+            'samtools_merge' in tool.id
```
nuwang, line 126: I was originally thinking that we would have two different users - one who would be exclusively routed to minio and the other to aws gateway. Was there a specific reason to choose this configuration?

cat-bro: yeah, I didn't have their email addresses so I thought I would use a test user as a first pass and add them later

jlqfab (approved):

PR #2203, opened 2024-09-26
install github hub package for jenkins server

jlqfab:
The github Hub package is now included in the standard Ubuntu Apt repos, installing via `common` role with `group_packages`.


cat-bro (approved):
Thanks @jlqfab !

PR #2210, opened 2024-10-02
make singularity volumes overridable context variables

cat-bro:
Rearrange slurm_singularity_volumes etc to be context variables on the default tool instead of hard coding them on destinations. This makes it possible for an aws gateway user to have '/mnt/aws_gateway:ro' appended to volume strings for their jobs.

@jlqfab @nuwang from a dry run this appears to just work.

### DIFF
--- a/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/users.yml.j2
+++ b/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/users.yml.j2
@@ -309,5 +309,8 @@ users:
       params:
         object_store_id: minio_test
     selenium_st_aws_gateway@genome.edu.au:
+      context:
+        slurm_singularity_volumes: "{{ slurm_singularity_volumes }},/mnt/aws_gateway:ro"
+        slurm_docker_volumes: "{{ slurm_docker_volumes }},/mnt/aws_gateway:ro"
       params:
         object_store_id: aws_gateway_test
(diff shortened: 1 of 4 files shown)


files/galaxy/dynamic_job_rules/production/total_perspective_vortex/users.yml.j2
```diff
@@ -309,5 +309,8 @@ users:
       params:
         object_store_id: minio_test
     selenium_st_aws_gateway@genome.edu.au:
+      context:
+        slurm_singularity_volumes: "{{ slurm_singularity_volumes }},/mnt/aws_gateway:ro"
```
nuwang, line 313: Why is this read-only though? Shouldn't it be writable for output files?

cat-bro: write is not required for any of the other user data folders. galaxy moves the files there after the job has run. see volumes listed in host_vars/galaxy

cat-bro:
If this is approved, please don't merge it. I'm confident that it works but would still like to keep an eye on galaxy when it goes in.

files/galaxy/dynamic_job_rules/production/total_perspective_vortex/users.yml.j2
nuwang: Right! I forgot the handler does the final move.

nuwang (approved):

PR #2214, opened 2024-10-03
add qld gpu pulsars to jwd deletion

cat-bro:


jlqfab (approved):
Thanks @cat-bro this will save us some headaches!

PR #2221, opened 2024-10-07
stacks denovomap memory

cat-bro:
update stacks_denovomap to have the same resources as stacks2_denovomap. This isn't configured to run on pulsar but it would be a good one for pulsar if it can run there


cat-bro:
this also updates all 'mem: 61.4' to 'mem: 62' because that is better use of the memory on the worker nodes

PR #2231, opened 2024-10-13
Deploy WALL-E to prod

neoformit:
Just tested this on dev.

- Deploy to dev, prod and staging
- Update playbooks to accommodate new "slack alerts" feature

This will require updating your `usegalaxy_eu.walle` role.

`requirements.yml` currently points to the `main` branch of the WALLE role. This should be updated with the next release (expecting 1.0.0): https://github.com/usegalaxy-eu/WallE


ansible.cfg
```diff
@@ -4,6 +4,7 @@ stdout_callback=debug
 stderr_callback=debug
 roles_path=roles
 collections_paths=collections
+private_key_file=~/.ssh/galaxy
```
cat-bro, line 7: Could you please remove this change - we have some VMs that need a different key so it's not convenient to declare the path in ansible.cfg

host_vars/staging.gvl.org.au.yml
```diff
@@ -335,3 +335,20 @@ toolmsg_messages:
       </a>
       to continue using this tool.
     class: warning
+
+walle_user_name: ubuntu
```
cat-bro, line 339: might be easier (less overhead) to leave this out of staging

dev_playbook.yml
```diff
@@ -122,3 +126,11 @@
       with_items:
         - "\t\tcp {{ nginx_log_olddir_root }}/* {{ nginx_long_term_log_dir }}/"
 
+    - name: Copy galaxy_jwd.py script to walle_script_location
```
cat-bro, line 129: why do this in dev playbook and not other playbooks?

files/galaxy/walle/test_object_store_conf.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0"?>
```
cat-bro, line 1: Could you please adapt the one from dev for this test file? This is so similar to production configuration that it is going to show up when searching for use of the string `/mnt/scratch` in the code base

host_vars/dev.gvl.org.au.yml
```diff
@@ -459,3 +459,20 @@ toolmsg_messages:
...
+walle_tool: interactive_tool
+walle_virtualenv: "{{ galaxy_venv_dir }}"
+walle_bashrc: "{{ galaxy_root }}/walle/.bashrc"
+walle_malware_database_location: "{{ galaxy_root }}/walle"
+walle_filesize_min: 0
+walle_verbose: false
+walle_kill: false
+walle_slack_alerts: true
+walle_slack_api_token: "{{ vault_walle_slack_api_token }}"
+walle_slack_channel_id: "{{ vault_walle_slack_channel_id }}"
```
cat-bro, line 474: channel_id does not need to be in the vault

ansible.cfg
neoformit: This will be overridden by the `--key-file` param - do those VMs use this, or do they default to an `id_rsa` key?

dev_playbook.yml
neoformit: Good catch - this is now included in walle role.

files/galaxy/walle/test_object_store_conf.xml
neoformit: Have just removed this - also now included in walle role.

ansible.cfg
cat-bro: I understand that but there are also contexts where key file does not have to be defined at all and the ssh key is already loaded (i.e. with jenkins) so pointing to a path that does not exist is inconvenient

neoformit: No worries, will remove now

cat-bro (approved):

PR #2264, opened 2024-10-28
Add ssh key for ubuntu user to be able to ssh to NCI workers more easily

cat-bro:
rename vault file ubuntu_maintenance_key to ssh_keys and add a key for internal ssh access between NCI pulsars and their workers.


cat-bro:
also marked w5 as inaccessible in the hosts file. It needs to be in the hosts file for the slurm role, but should not be in the pulsar_nci_training_workers group for now.

cat-bro:
NCI has deleted w5 so when this is rebuilt there will be 12 workers again.

I'll like to merge this now because I'll use the same pulsar_hop_key ssh key for pulsar-mel3

jlqfab:
Test if github is still giving errors.

jlqfab (approved):
LGTM!

PR #2267, opened 2024-10-30
put citations_export_message_html in galaxy.yml

cat-bro:
Minh has added an option for us to customise the 'cite galaxy' message which allows us to alter the text and/or the URL it links to. The galaxyhub URL it links to by default seems like the right URL, since it comprehensively describes how to cite many components of Galaxy. This pull request changes our text to say 'cite Galaxy Australia' instead of 'cite Galaxy' and leaves the link as is.


cat-bro:
the original PR for au branch was here: https://github.com/usegalaxy-au/galaxy/pull/2/files

PR #2274, opened 2024-11-04
AAGI tools

neoformit:
Minor update to deploy another AAGI tool on dev

### DIFF
--- a/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/dev_tools.yml.j2
+++ b/files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/dev_tools.yml.j2
@@ -1,25 +1,5 @@
 tools:
-  aagi_001:
-    cores: 4
-    mem: 8
-    env:
-      DOCKERHUB_USERNAME: aagisupport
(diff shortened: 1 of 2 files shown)


files/galaxy/dynamic_job_rules/dev/total_perspective_vortex/dev_tools.yml.j2
```diff
@@ -1,25 +1,5 @@
...
-    - if: |
-        not user or not 'aagi' in [role.name for role in user.all_roles() if not role.deleted]
-      fail: |
-        Access to this tool is restricted. Please submit a support ticket if you believe you should have access.
-    scheduling:
-      require:
-        - slurm
-
-  aagi_002:
+  aagi_*:
```
cat-bro, line 2: The above key probably works because substrings match, but `aagi_.*` (with dot before star) is more correct, as it is a python regex match, not glob

neoformit: Ahh thank you for that!

cat-bro (approved):

PR #2275, opened 2024-11-05
take pulsar-qld-gpu4 and gpu5 offline for GPU3 replacement

jlqfab:
To be merged prior to hypervisor switch off at 2pm Wed 6th Nov.


cat-bro:
Could you please also add `pulsar-qld-gpu-other` to `accept` tags on gpu1 and gpu2 in destinations.yml.j2 with a comment that this is temporary while 4 and 5 are offline? This means that dorado and helixir jobs will queue there instead. Even if there is no room for them to run it's better than them failing on the spot, and they can be rescheduled later.

cat-bro (approved):

PR #2276, opened 2024-11-05
reduce pulsar QLD capacity in tpv

cat-bro:
temporarily reduce tpv's understanding of resources available at pulsar-QLD


jlqfab (approved):
Thanks @cat-bro !

PR #2285, opened 2024-11-13
disable singularity for retrieve_scxa

jlqfab:
Fix issue for EBI SCXA Data Retrieval where get the following error: `wget: unable to resolve host address 'ftp.ebi.ac.uk'`. This appears to be caused by running the tool within singularity.


cat-bro (approved):
Thanks @jlqfab !

PR #2295, opened 2024-11-22
Update user_preferences_extra_conf.yml.j2

mthang:
Add Xnat account section to user preference

### DIFF
--- a/templates/galaxy/config/user_preferences_extra_conf.yml.j2
+++ b/templates/galaxy/config/user_preferences_extra_conf.yml.j2
@@ -90,7 +90,8 @@ preferences:
               label: App Password
               type: password
               required: False
-    
+
+{% if inventory_hostname.startswith('dev') %}
     xnat_account:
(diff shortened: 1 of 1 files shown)


templates/galaxy/config/user_preferences_extra_conf.yml.j2
```diff
@@ -90,4 +90,22 @@ preferences:
               label: App Password
               type: password
               required: False
+    
+    xnat_account:
```
cat-bro, line 94: Could you please surround this block with {% if %} tags in the same way the the chatgpt block is surrounded earlier in the file? This way it will only be on dev while it's a dev tool

mthang: @cat-bro I just added the (% if %} to the user_preferences.

cat-bro (approved):
Thanks @mthang

PR #2303, opened 2024-11-28
unnest object store folders

cat-bro:
Some of the interstate folders are nested. It would be better if (a) no object store folder is within another object store folder and (b) an object store folder is within a volume rather than being the volume

This needs to be merged/deployed when there are not running jobs using QLD/WA data, folders moved and paths carefully checked


jlqfab (approved):
looks sensible to me!

PR #2321, opened 2024-12-16
prepare pmel3 for shutdown

cat-bro:
Allow pulsar-mel2 to run user jobs again, and add 'require: pulsar-quick' to pulsar-mel3 so that it can only run jobs from tools with that tag.


cat-bro:
There is no need to shut it down yet - it has some long jobs running. I'd like to replace it with fresh ubuntu 22.04 VMs in January

jlqfab (approved):

PR #2328, opened 2024-12-20
Resubmittable tool in dev tpv

cat-bro:
Add an abstract tool `__resubmittable_tool` to dev's vortex config. I've been trying out the resubmit option and I can only get it to resubmit once - from reading issues/PRs on TPV's github I think others have come across this.

Even so, the ability to automatically resubmit a job once would be beneficial in the case of tools with very unpredictable memory requirements.


cat-bro:
Pinging @nuwang but no rush - I'll just merge this because it's dev

cat-bro:
superseding this on with #3386

PR #2331, opened 2025-01-06
jenkins script to check for duplicate data table entries

cat-bro:
If a data table is added to cvmfs with an entry that matches an existing one from GA's custom indices data, jobs will fail and users won't report it for ages. This has happened recently with interproscan, eggnog and bwameth.

Jenkins can run this script every day and post an alert if there are dupicate keys.


jlqfab (approved):
LGTM!

PR #2335, opened 2025-01-09
Update tools.yml

igormakunin2:
limit for fastani: four concurrent jobs per user was added


cat-bro (approved):
Thanks @igormakunin2 !

PR #2337, opened 2025-01-09
rebuild pulsar-mel3

cat-bro:
Pulsar-mel3 cluster with a new VM for NFS. The original volume has been imported so that we don't have to rsync the custom indices data again. The pulsar app and files will be on a fresh volume.


jlqfab (approved):
LGTM

jlqfab (approved):
Looks good to go as far as I can see.

PR #2347, opened 2025-01-17
set default for variable

cat-bro:
quick fix


cat-bro:
this is lazy but so is the existence of the variable 'is_galaxy_head_node' which I'd like to factor out. For the moment, can accumulate debt

jlqfab (approved):

PR #2352, opened 2025-01-23
fix update-roles bug

cat-bro:
this has been bugging out when it encounters a new role in requirements.yml


jlqfab (approved):
Good catch!

PR #2362, opened 2025-01-29
extra environment variable for pilon

cat-bro:
Java options might not be being passed through the normal `_JAVA_OPTIONS` environment variable.

I doubt this will fix pilon but this should be tried first


jlqfab (approved):
worth a try!

PR #2363, opened 2025-01-30
All tools can have _JAVA_OPTIONS

cat-bro:
a bunch of pilon jobs failed because the tool had the `_JAVA_OPTIONS` environment variable but not the `SINGULARITYENV__JAVA_OPTIONS` environment variable.

All tools can have java options set - they won't be used if it's not a java tool anyway. The lack of propagation of regular environment variables for container jobs is annoying but easy enough to get around if the variable can be put on the default tool like this.


jlqfab (approved):
Looks good!

jlqfab:
Actually, just realised this will set `-Xmx` to the default memory, won't it?

PR #2387, opened 2025-02-07
Move subdomains switcher function to script.js.j2 template

neoformit:
This should fix the site switcher on 24.2 - still not clear to me why but it seems like the `function` property in the webhooks `config.yml` no longer works, though it is still used in some of the [demo webhooks](https://github.com/galaxyproject/galaxy/blob/dev/config/plugins/webhooks/demo/tool_list/config.yml#L9). It wouldn't even do a `console.log("HELLO WORLD")` in my testing. Maybe worth pinging matrix or galaxyproject github to see if they're aware?

FYI I tested this on 24.2 only but I think it should be backwards compatible.


cat-bro:
LGTM and sounds like it would be worth letting galaxy know about the demo webhooks.

cat-bro (approved):

PR #2393, opened 2025-02-10
Switch toolmsg to type 'onload'

neoformit:
This one was easier - it still works on 24.2 but I changed it to `type: onload` to get rid of the blank button in the masthead.


cat-bro (approved):
Thank you @neoformit !

PR #2402, opened 2025-02-12
Version webhooks for galaxy 24.2

neoformit:
I think this should allow us to version webhooks more easily? Also good for other Galaxies that have used these webhooks, so they can clearly see the difference here.


cat-bro:
So @neoformit I can merge this, then tomorrow change webhooks_plugins on production to the _24.2 suffix ones?

neoformit:
Yes, I think that should work!

cat-bro (approved):

PR #2411, opened 2025-02-17
Fix subdomain name typo

neoformit:
Sorry, I called it `microgalaxy` instead of `microbe`


cat-bro:
could you please also add this to host_vars/dev.gvl.org.au.yml certbot_domains?

cat-bro (approved):

PR #2449, opened 2025-03-06
add data27

cat-bro:
Create a data27 directory and remove explicit setting of job_working_directory since the default is now the right one in every case.


cat-bro:
For posterity, this PR was once going to create two object store backends for data27: one for slurm jobs with normal job working directory, one for pulsar jobs with job working directory on the user disk.

I'm abandoning that because it puts a pattern in place that is going to make it much harder to bring in scratch quotas. It would be preferable to be able to override `job_working_directory` in TPV regardless of where the user data was being written to.

This would be ideal:
```
# separate galaxy location for pulsar job JWDs close to upload store
galaxy_pulsar_job_working_directory: "{{ upload_store_volume_dir }}/pulsar_jwds"

host_galaxy_extra_dirs:
  - "{{ galaxy_pulsar_job_working_directory }}"
  ```

PR #2454, opened 2025-03-10
Fix gms postgres

neoformit:
I found some issue with the postgresql config that needed fixing as this broke gms-dev.

- `postgresql_listen_addresses` is not used - needs to be set in `postgresql_global_config_options`
- The `postgresql_hba_entries` prevented local/peer connection from `postgres` user


cat-bro (approved):
Thanks @neoformit!

PR #2461, opened 2025-03-13
remove pulsar-quick tag from shovill

cat-bro:


cat-bro:
it is not being quick enough

jlqfab (approved):

PR #2465, opened 2025-03-13
Simplify blast variables, update roles and add new local roles

cat-bro:


cat-bro:
There are things I can apply across the board here like pulsar pre and post tasks, the tmp-relocation role etc. I think this is the list of roles, in the right order, for any of the production pulsar head nodes that are on ubuntu 22.04

jlqfab (approved):
LGTM

PR #2467, opened 2025-03-13
Set concurrent user limit for trinity

cat-bro:
This is currently set for `/trinity.*` and not `/trinity/trinity/.*`. The simplest thing to hope for is that inheritance hasn't worked as expected. If this doesn't fix it, it's a much bigger problem.


cat-bro:
I think the inheritance does work. Will revisit if this becomes a problem again.

PR #2475, opened 2025-03-18
really big cvmfs caches for high mem and blast

cat-bro:
even with 250GB of cache space available, pqhm1 is cleaning out its cache several times a day. Making space for this seems more pressing than space for files. If file space becomes an issue this can be revisited.

### DIFF
--- a/host_vars/pulsar-high-mem1/pulsar-high-mem1.yml
+++ b/host_vars/pulsar-high-mem1/pulsar-high-mem1.yml
@@ -24,7 +24,7 @@ use_internal_ips: false
 
 # cvmfs
 cvmfs_cache_base: /mnt/var/lib/cvmfs
-cvmfs_quota_limit: 250000
+cvmfs_quota_limit: 819200
 
 #Monitoring for Staging. Once all VM monitoring has moved to stats.usegalaxy.org.au, this can be put in all.yml and removed here.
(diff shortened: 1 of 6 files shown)


host_vars/pulsar-high-mem1/pulsar-high-mem1.yml
```diff
@@ -24,7 +24,7 @@ use_internal_ips: false
 
 # cvmfs
 cvmfs_cache_base: /mnt/var/lib/cvmfs
-cvmfs_quota_limit: 250000
+cvmfs_quota_limit: 819200
```
jlqfab, line 27: Where does this 'magic' number come from?

cat-bro: 800 times 1024

jlqfab: more than tripling in size from previously

cat-bro: Yes. There is a 10TB disk that is rarely more than half full, and a lot of large databases in cvmfs that are being accessed for kraken2 jobs.  This seems like good use of the space.

jlqfab (approved):

PR #2476, opened 2025-03-18
allow extra arguments to make run with a catch

cat-bro:
With this change you could run
`make run dev -- --skip-tags users`

or any number of ansible-playbook command flags: --extra-vars etc

note the extra `--` between the original command and the extra flags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,26 +1,32 @@
 # Management commands
 
-
-# This 'make' hack allows us to accept a single positional argument for the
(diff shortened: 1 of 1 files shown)


cat-bro:
I don't love this because it uses made-up logic, but it is convenient

neoformit:
Should we put a few lines in the README on running with `make`?

neoformit (approved):

PR #2482, opened 2025-03-21
disable unattended-upgrade on pulsar-qld-gpu VMs

jlqfab:
`unattended-upgrades` runs in the background and by default does automatic upgrades. On pulsar-qld-gpu VMs which run alphafold etc, when the kernel is updated the Nvidia drivers become out of sync with kernel version resulting in `nvidia-smi` giving the error `Failed to initialize NVML: Driver/library version mismatch`.

The solution is to configure `unattended-upgrades` via `/etc/apt/apt.conf.d/20auto-upgrades` so that it only updates package lists but doesn't upgrade the packages. This means that can still see when updates are available, with `apt list --upgradable`, and upgrades can then be applied manually when convenient.


jlqfab:
This has been tested on pulsar-qld-gpu-dev with: `ansible-playbook -i hosts --key-file ~/.ssh/galaxy one-offs/initial_nvidia_config_disable_unattended-upgrade_and_disable_cgroups_v2_playbook.yml --limit pulsar-qld-gpu-dev --tags disable_upgrade`

cat-bro (approved):

cat-bro:
LGTM!

PR #2483, opened 2025-03-21
Flush handlers to prompt postgres restart

neoformit:
This seems to be the proper fix for a recent workaround that I made. Tested and working.

### DIFF
--- a/roles/galaxy_media_site/tasks/main.yml
+++ b/roles/galaxy_media_site/tasks/main.yml
@@ -1,5 +1,9 @@
 ---
 
+- name: Ensure PostgreSQL is restarted immediately after configuration change
+  meta: flush_handlers
+  tags: postgresql
+
 - name: Add the deadsnakes PPA repository
   apt_repository:
     repo: ppa:deadsnakes/ppa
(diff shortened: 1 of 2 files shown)


roles/galaxy_media_site/tasks/main.yml
```diff
@@ -1,5 +1,9 @@
 ---
 
+- name: Ensure PostgreSQL is restarted immediately after configuration change
+  meta: flush_handlers
+  tags: postgresql
```
jlqfab, line 5: `flush_handlers` only executes handlers that have been notified up to this point. So this should be added _after_ the changes, where you originally had the `restart postgresql` task

jlqfab: see: [ansible.builtin.meta module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/meta_module.html)

neoformit: Yes, the idea is that it flushes handlers which were notified in the postgresql role which runs before this.

PR #2501, opened 2025-03-28
add a slurm partition for msconvert-friendly workers

cat-bro:
For the older version of msconvert on GA (3.0.20287.2), jobs are failing on workers 4-6 because the .wine directories have 1777 perms instead of 0700.


cat-bro:
workers 4-6 are the ones that ITs run on. It's very mysterious and unsatisfactory.

PR #2529, opened 2025-04-11
add salmon to dev tpv rules

cat-bro:


cat-bro:
Same settings as production for someone who wants to test it with/without data  tables

PR #2589, opened 2025-05-14
reduce retention_disk variable in galaxy db backup from 7 to 5

cat-bro:
Every morning usage of `/mnt` on galaxy-backup spikes above 90%, let’s keep fewer of these on disk. JL is this the right way?


jlqfab (approved):
that should do it.

PR #2590, opened 2025-05-14
bigger cvmfs caches for small workers

cat-bro:
Small/medium workers that have 200GB mounted disks can use 100GB for cvmfs cache


jlqfab (approved):
LGTM - ignoring typo in your comment (small should be 20G not 200G)

PR #2593, opened 2025-05-15
use galaxy main kraken2 rule on production

cat-bro:
Galaxy Main’s kraken2 rule works well on dev. I was initially worried about frequent file size lookups but apparently this is not a super expensive operation, though it will be done over and over.

This also adds cmvfs_cache_100plus and cvmfs_cache_800plus tags to various destinations so that bakta and kraken2 can prefer destinations with non-tiny caches.

### DIFF
--- a/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/destinations.yml.j2
+++ b/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/destinations.yml.j2
@@ -147,6 +147,7 @@ destinations:
     scheduling:
       accept:
         - pulsar-mel3
+        - cvmfs_cache_100plus
         - bakta_database
         - funannotate
         - eggnog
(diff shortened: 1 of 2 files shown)


cat-bro:
Reading more, I am not sure about this. In conjunction with a concerrency rule this could be working too hard. This is Galaxy Main’s whole entry for kraken2 and they have commented out the concurrency rule for kraken2 https://github.com/galaxyproject/usegalaxy-playbook/blob/main/env/common/templates/galaxy/config/tpv/tools.yaml.j2#L870-L911

They still have other concurrency rules which makes me think this particular rule applied over and over is problematic.

It would be better (and not much work) to compose a dictionary of table keys and file sizes and just put it in the ‘mem’ block to avoid lots and lots of os.path.getsize() lookups.

nuwang (approved):
That sounds good to me. Let's just merge and see what the performance impact is like. We should be able to judge by log timestamps how long it takes to evaluate each loop (or alternatively, maybe use the timeit module for a more accurate measurement?)

PR #2608, opened 2025-05-28
Update local_tool_conf.xml

mthang:
add dorado error correct wrapper to Production


cat-bro (approved):

cat-bro:
@mthang this should be visible and usable by 14:30

PR #2616, opened 2025-06-03
Fix bug in toolmsg plugin

neoformit:
Stupid syntax error, will need to let EU/main know about this (I can't remember if main is also using this)


cat-bro (approved):

neoformit:
Thanks @cat-bro can you please let me know when this is deployed so I can test?

neoformit:
Thanks - looks like it's fixed. Thanks for raising this.

sanjaysrikakulam:
@neoformit Sorry, I didn't notice this when I deployed it. In the EU, I added an [extra if condition](https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/templates/galaxy/webhooks/toolmsg_24.2/script.js.j2#L9-L17), so I seem to have fixed this syntax issue for the EU when I deployed it.

neoformit:
Thanks @sanjaysrikakulam you're on the ball! Yeah this error is in the Ansible template so does not come up in my local testing. Good that you ended up fixing it - I would have expected an issue from you sooner otherwise.

PR #2622, opened 2025-06-05
do not use replication slot on galaxy db

cat-bro:
Using a replication slot runs the risk of db disk at ETCA filling up with WALs if the replication db stalls. Without the replication slot there is a risk of the replica falling behind, but this seems like good trade-off.


jlqfab (approved):
LGTM

PR #2645, opened 2025-06-13
Update the QLD high memory pulsars to 22.04 and modern roles

cat-bro:


jlqfab (approved):
I love orthogonality! LGTM!

PR #2649, opened 2025-06-19
picard tmp

cat-bro:
picard does not honour ‘tmp_dir: true’, apparently it doesn’t respect the $TMP_DIR environment variable and JAVA_OPTIONS is the way to go. Should resolve issues from the past few days with worker disks filling up.


jlqfab (approved):
Good catch @cat-bro !

cat-bro:
the change is using the default JAVA_OPTIONS from the default tool and using the `-Djava.io.tmpdir` copied from Galaxy Main’s TPV

PR #2651, opened 2025-06-19
Some TPV rules, high-mem for masurca

cat-bro:
Set singularity enabled for bgruening tools from A to Canu.
Move ‘minimum_singularity_version’ rule to the default tool. There are many occasions where older versions of tools need to keep running with conda envs.
Set mem much higher for masurca_simple because of OOM errors.

### DIFF
--- a/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/default_tool.yml.j2
+++ b/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/default_tool.yml.j2
@@ -24,6 +24,9 @@ tools:
       pulsar_score_slurm_max_cores: 12  # if cores >= this number, do not set preference for slurm
       pulsar_score_slurm_max_mem: 48  # if mem >= this number, do not set preference for slurm
  
+      # resolver selection
+      minimum_singularity_version: null
+
       # singularity and docker volumes
       slurm_singularity_volumes: "{{ slurm_singularity_volumes }}"
       slurm_docker_volumes: "{{ slurm_docker_volumes }}"
@@ -36,6 +39,14 @@ tools:
     - id: login_required_rule
       if: require_login and user is None
       fail: 'Please log in to use {tool.id}'
+    - id: minimum_singularity_version_positive_rule
+      if: minimum_singularity_version is not None and helpers.tool_version_gte(tool, minimum_singularity_version)
+      params:
+        singularity_enabled: true
+    - id: minimum_singularity_version_negative_rule
+      if: minimum_singularity_version is not None and helpers.tool_version_lt(tool, minimum_singularity_version)
+      params:
+        singularity_enabled: false
     - id: max_concurrent_job_count_for_tool_rule
       if: |
         total_limit_exceeded = False
--- a/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
+++ b/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
@@ -45,6 +45,9 @@ tools:
   toolshed.g2.bx.psu.edu/repos/bgruening/10x_bamtofastq/10x_bamtofastq/.*:
     params:
       singularity_enabled: true
+  toolshed.g2.bx.psu.edu/repos/bgruening/add_line_to_file/add_line_to_file/.*:
+    params:
+      singularity_enabled: true
   toolshed.g2.bx.psu.edu/repos/bgruening/agat/agat/.*:
     params:
       singularity_enabled: true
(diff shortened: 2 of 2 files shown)


files/galaxy/dynamic_job_rules/production/total_perspective_vortex/default_tool.yml.j2
```diff
@@ -36,6 +39,14 @@ tools:
...
       if: require_login and user is None
       fail: 'Please log in to use {tool.id}'
+    - id: minimum_singularity_version_positive_rule
+      if: minimum_singularity_version is not None and helpers.tool_version_gte(tool, minimum_singularity_version)
+      params:
+        singularity_enabled: true
+    - id: minimum_singularity_version_negative_rule
+      if: minimum_singularity_version is not None and helpers.tool_version_lt(tool, minimum_singularity_version)
+      params:
+        singularity_enabled: false
```
nuwang, line 49:
```suggestion
    - id: minimum_singularity_version_rule
      if: True
      params:
        singularity_enabled: {'true' if minimum_singularity_version is not None and helpers.tool_version_gte(tool, minimum_singularity_version) else 'false'}
```

cat-bro: Interestingly, this doesn’t work:
```
      params:
        singularity_enabled: {"true" if minimum_singularity_version is not None and helpers.tool_version_gte(tool, minimum_singularity_version) else "false”}
``` 

The error is

```
ruamel.yaml.parser.ParserError: while parsing a flow mapping
  in "tpv_check/total_perspective_vortex/default_tool.yml", line 45, column 30
expected ',' or '}', but got '<scalar>’
```

I tried a few variations. Some didn’t raise errors but were always truthy. The only one I can find that works is this:

`singularity_enabled: '{str(helpers.tool_version_gte(tool, minimum_singularity_version)).lower()}’`

This works but it has the knock-on effect of causing a bunch of parameters to be defined that wouldn’t be otherwise because of destination rules that look at whether singularity_enabled is “true” or “false” here https://github.com/usegalaxy-au/infrastructure/blob/master/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/destinations.yml.j2#L20-L21

I think this needs to stay as it is with two rules.

nuwang: Isn't it because the braces are interpreted as YAML dicts? Enclosing the braces in doubles quotes should work?:
```yaml
      params:
        singularity_enabled: "{'true' if minimum_singularity_version is not None and helpers.tool_version_gte(tool, minimum_singularity_version) else 'false'}"
```

nuwang: Another issue here is that with type checking, params are always treated as strings (as they should be). If we want actual booleans for decision making, we should use context variables I think.

cat-bro: I don’t want to set ‘if: True’ and apply this to every tool even when minimum_singularity_version has not been set, because we may have already set 'singularity_enabled: true’ on the tool and I don’t want to flip that to false just because a minumum version did not need to be defined for that particular tool.

Having an expression that evaluates to a string is workable (as my example above), but it does muck up the logic around setting parameters in destinations.yml.j2.

This all needs a more comprehensive solution, but merging this PR as is with the positive rule and negative rule will make it easier to enable singularity for tools with many versions, and default to enabling singularity for toolshed tools which would save a lot of time.

cat-bro: Or can I somehow evaluate the expression here? https://github.com/usegalaxy-au/infrastructure/blob/master/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/destinations.yml.j2#L21? This is where setting it as an expression is difficult - as soon as `singularity_enabled` is an expression, this ‘if’ is always truthy.

nuwang: On the whole, this has opened up an interesting set of issues.

I believe that params actually being a bool is a type error that should get caught by the new linter. Instead, it should be a string. If it's a non-empty string, it will always be truthy after evaluation. So this is also a test of the new linter in a way.

Perhaps we could do something like this instead:
```
if: entity.params.get('singularity_enabled') and entity.params.get('singularity_enabled') != 'false'
```

or use the `util.asbool` function from galaxy, which does all that automatically and produces an actual bool value:
```
if: util.asbool(entity.params.get('singularity_enabled'))
```

If this is needed urgently though, sure, maybe merge as is, and we can sort this out later, but I think the above option should work.

PR #2656, opened 2025-06-23
Update tools.yml

mthang:
remove dorado_error_correct entry (this is a local copy of the tool) from tpv for now. This tool has been pushed to the toolshed and will install this tool from the galaxy toolshed. The new entry will be added after the installation.


cat-bro:
Hi @mthang instead of removing this, could you please change the key to the new ID?

cat-bro (approved):
Thanks @mthang!

PR #2658, opened 2025-06-23
spring cleaning and better exporting

cat-bro:
Remove default nfs_exports from pulsarservers.yml.

In production variables, replace `*` with something less inclusive for nfs_exports.

Remove pawsey-user-nfs files and hosts entry.

Some arbitrary reordering of pulsar-nci-training files.


jlqfab (approved):
Looks good!

PR #2659, opened 2025-06-23
Update destination scoring

cat-bro:
The local TPV ranking function looks at the resources used as a proportion of total resources at a destination. When the slurm-training destination is in use the jobs do not count towards the slurm destination’s resource use but they should. If the training jobs are not accounted for, the slurm destination appears to be be more available than it is, so it gets too many jobs.


jlqfab (approved):
Looks good to me!

PR #2667, opened 2025-07-02
More RAM for large RagTag alignment

TomHarrop:
Hello! This is a non-urgent request for more RAM for large RagTag jobs.

I tried to do a whole genome alignment with two genomes of around 2 Gigabases each and ran out of RAM (GA job 4043685).

I don't know how much RAM is enough, I haven't tested it locally  

Thanks :)


cat-bro (approved):
Thank you @TomHarrop !

PR #2669, opened 2025-07-02
cvmfs singularity for all pulsars

cat-bro:
So far, the jobs are alright and the runtimes are slightly faster if the container is cached. It seems like a good idea to start using this with all pulsars immediately.


jlqfab (approved):
Go for it!

PR #2674, opened 2025-07-06
restrict DATA_FETCH to handler_0

cat-bro:
Whatever is holding the handlers up apparently comes from DATA_FETCH jobs.  These have been restricted to run on handler_0 only which simplifies things a bit. handler_0 is being restarted regularly and the other handlers are fine now.


cat-bro:
ref https://github.com/galaxyproject/galaxy/issues/20600

PR #2675, opened 2025-07-07
reduce wait time for restricted interactive tools

cat-bro:


cat-bro:
I think this has achieved what was intended, and potentially could have the same effect while being less punishing to new good users

jlqfab (approved):

PR #2677, opened 2025-07-08
give test user ability to test tools without singularity

cat-bro:
Test users can put ‘singularity’ in their history tags to run tools with singularity even if they are not configured to do so. There is no equivalent for disabling singularity and this would be useful.


jlqfab (approved):
Give it a go!

PR #2698, opened 2025-07-23
Add logrotate tasks to common role

cat-bro:
Add tasks to the common role to update the frequency of log rotation of syslog. Root disks on galaxy/galaxy-handlers have too much space taken up by uncompressed syslog files because logs are rotated weekly. Switch this on on dev first to test it.

Put dev onto release_25.0_au_dev branch which is the same as release_25.0 but will need to be rebased regularly.

Switch on daily cron jobs to clean up tmpdisk folders. Keep it switched off on galaxy workers for now.

### DIFF
--- a/host_vars/dev.gvl.org.au.yml
+++ b/host_vars/dev.gvl.org.au.yml
@@ -2,8 +2,7 @@
 
 add_hosts_handlers: yes
 
-# Login override - checksum for Login.vue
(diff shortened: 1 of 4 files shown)


jlqfab (approved):
Looks good. But I can't see anything that does this: "Switch on daily cron jobs to clean up tmpdisk folders." as per your comment.

jlqfab (approved):

PR #2699, opened 2025-07-23
take overridden rules out of destinations.yml.j2

cat-bro:
Overriding the pulsar_docker_volumes context variable should be enough without overriding rules, if it isn’t I will revert this.


nuwang (approved):
This lgtm!

PR #2726, opened 2025-08-14
Try to fix ENA upload toolmsg... again!

neoformit:
Still not working on prod - it looks like it must load and unload the `tool-card-body` element multiple times (Vue re-renders probably). Hopefully this fix can handle it... we'll see.


jlqfab:
webhooks role failed so it wouldn't have updated

neoformit:
Ok thanks, will get onto that this arvo. Looks like it's just missing a mkdir...

PR #197, opened 2021-03-25
change __galaxy_config to host_galaxy_config, galaxy_config to group_galaxy_config and combine the two

cat-bro:
Fix the issue with galaxy_config across the vars files.

Marking this as draft so I can test it out more tomorrow

PR #2749, opened 2025-09-02
workaround for broken apt-key in dj-wasabi.telegraf

jlqfab:
On new VMs (still ubuntu 22.04) the `dj-wasabi.telegraf` role breaks due to the repo key not being dearmored. Existing VMs that have already added the telegraf repo don't appear to be affected, possibly due to the key already existing in the global keyring `/etc/apt/trusted.gpg`.

This workaround does what the `dj-wasabi.telegraf` role should do: download and dearmor the key (now to `/etc/apt/keyrings/` instead of `/usr/share/keyrings/` as recommended best practice). It then passes the location of the dearmored repo key via a variable `telegraf_repository_params`.

This approach has been manually tested on the new galaxy workers. This serves to make the workaround more permanent and documented.

@cat-bro For noting on new VMs.

### DIFF
--- a/galaxy-workers_playbook.yml
+++ b/galaxy-workers_playbook.yml
@@ -18,6 +18,44 @@
           path: "{{ item }}"
         with_items:
           - "{{ qld_file_mounts_path }}"
+
+      # the following is a workaround for dj-wasabi.telegraf role breaking on new VMs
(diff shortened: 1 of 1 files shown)


jlqfab:
Submitted PR: Debian/Ubuntu: make telegraf_repository_params overridable #197
https://github.com/dj-wasabi/ansible-telegraf/pull/197

PR #2767, opened 2025-09-14
oom updates 2025-09-15

cat-bro:


nuwang (approved):
Also transfer to shared db?

cat-bro:
Thanks @nuwang. I have a script that I run to put local changes into a shared db PR. I can adapt this to look for instances where our mem is higher.

Main has been adding mem as linear in terms of input size  for some tools https://github.com/galaxyproject/usegalaxy-playbook/blob/main/env/common/templates/galaxy/config/tpv/tools.yaml.j2#L684-L686 and these will great for shared db.

PR #2798, opened 2025-09-30
production to use latest tpv shared db

cat-bro:
Versioning has been introduced in the tpv shared db https://github.com/galaxyproject/tpv-shared-database?tab=readme-ov-file#version-guide.

I’d be happy to use the latest updates to shared-db as they come in and revert to a version if this turns out to be a problem.


nuwang (commented):
We could also go for `db-v2.yml`. That's an alias for latest at the moment, but we are less likely to risk accidental breakage if a v3 is released.

PR #2820, opened 2025-10-07
remove insppb.hostname role

cat-bro:
This role causes anomalies in the `/etc/hosts` file. JL has looked into this and concluded that it is not needed.


jlqfab (approved):
bye bye! :-)

PR #2835, opened 2025-10-14
one off fix etc hosts

cat-bro:
Revert `/etc/hosts` files corrupted by insspb.hostname

The host here is dev_group but I’ll manually change it to run on one group at a time. There is a debug task here that shows line by line changes.


jlqfab (approved):
Looks good! Thanks @cat-bro

PR #2836, opened 2025-10-15
Update dev_destinations.yml.j2 for tpv 3.1.1

cat-bro:
rule inheritance breaks TPV because the pydantic model expects every member of `rules` to have an `if`.

This is probably not a bug but a change in behaviour that needs to be accommodated.


cat-bro:
ref https://github.com/galaxyproject/total-perspective-vortex/issues/159

PR #2838, opened 2025-10-15
remove conda rule for snpeff v5.2

cat-bro:
Galaxy Europe fixed the recipe and the container for this.


cat-bro:
A few lines above, singularity_enabled: true has been set for the whole snpEff family

PR #2842, opened 2025-10-17
Revert "update pulsar score slurm threshold"

cat-bro:
Reverts usegalaxy-au/infrastructure#2809


cat-bro:
There have been a few job failures due to galaxy-handlers being so overloaded that pulsar cannot post files back. In future I wonder if it would be possible to make scheduling decisions based on the system load of galaxy-handlers (the scheduler runs on that VM) but for the moment it makes sense just to revert this.

jlqfab (approved):

PR #2853, opened 2025-10-24
back up venv list

cat-bro:
Galaxy Australia has a private github repo containing useful backups: files managed by galaxy, tiaas and history mailer dbs, probably should contain media site db as well.

This is to save a daily copy of pip list so it is easy to see when packages change.

### DIFF
--- a/jenkins/config_backup/config_backup_playbook.yml
+++ b/jenkins/config_backup/config_backup_playbook.yml
@@ -41,10 +41,10 @@
         command: /mnt/galaxy/venv/bin/pip freeze
         register: pip_packages
       - name: Save package list locally
-        local_action:
-          module: copy
+        copy:
           content: "{{ pip_packages.stdout }}"
           dest: "{{ config_backup_dir }}/galaxy_venv/venv_packages.txt"
+        delegate_to: localhost
 - hosts:
     - galaxy-backup
   become: true


jenkins/config_backup/config_backup_playbook.yml
```diff
@@ -34,6 +35,16 @@
         dest: "{{ config_backup_dir }}/tiaas/"
       with_items:
         - db.sqlite3
+    - name: Copy list of venv packages
+      block:
+      - name: Get list of packages
+        command: /mnt/galaxy/venv/bin/pip list --format=freeze
```
jlqfab: @cat-bro would `command: /mnt/galaxy/venv/bin/pip freeze` be more suitable for reproducibility?

jenkins/config_backup/config_backup_playbook.yml
```diff
@@ -34,6 +35,16 @@
...
+        register: pip_packages
+      - name: Save package list locally
+        local_action:
```
jlqfab: @cat-bro optionally, you can use `delegate_to: localhost` instead of the older form `local_action`. They are both equivalent, so it's ok as is.

jenkins/config_backup/config_backup_playbook.yml
cat-bro: yes indeed! thank you

jlqfab (approved):

PR #2876, opened 2025-11-07
fix volumes string for data managers

cat-bro:
singularity_volumes for data managers has been overridden to allow /mnt/custom-indices to be `rw`. Remove /mnt/user-data-volC and /mnt/user-data-pawsey from this list.


jlqfab (approved):
Good catch!

PR #2879, opened 2025-11-13
remove webdav task

cat-bro:
GA’s playbook uninstall webdav and install it at a specific commit from 2022: https://github.com/ezhov-evgeny/webdav-client-python-3/commit/0f17fa7946e66f7963db367d0d6b2e7f940ebeb8. Galaxy’s conditional requirements now also use a specific commit of webdav (from 2023) https://github.com/ezhov-evgeny/webdav-client-python-3/commit/98c23d1abd15efc3db9cfc756429f00041578bc2

so we can probably use the version from conditional-requirements and stop uninstalling/reinstalling webdav.

### DIFF
--- a/dev_playbook.yml
+++ b/dev_playbook.yml
@@ -96,15 +96,6 @@
     - name: Install slurm-drmaa
       package:
         name: slurm-drmaa1
-    - name: Uninstall the existing web-dav-client
-      pip:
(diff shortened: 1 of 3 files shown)


nuwang (approved):
lgtm!

PR #2880, opened 2025-11-13
Add bring-your-own-file-sources to production

cat-bro:
There are useful things here that people can already use if they have access. Unfortunately the generic S3 template does not work for my university S3 storage because it is finicky about how the data is read and written. We may have the option of developing something specific  for this in the future.

In the meantime this should be enabled.

This PR also adds the additional venv packages for boto3. There may be other packages that need to be installed for other file source types - cross that bridge when we come to it.

### DIFF
--- a/group_vars/galaxyservers.yml
+++ b/group_vars/galaxyservers.yml
@@ -145,6 +145,12 @@ tpv_packages: |
 
 additional_packages:
   - flower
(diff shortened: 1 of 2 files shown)


cat-bro:
If reviewing, please do not merge. Might need extra OK from the product owner.

jlqfab (approved):
LGTM

PR #2894, opened 2025-11-24
stats-2025 grafana server inventory and host_vars

jlqfab:
Definitions for the new Grafana server on Ubuntu 24.04.

Notes:
- the playbook to install the stats-2025 server is `stats_playbook.yml` and doesn't need any changes it just needs to be run with `--limit stats-2025.usegalaxy.org.au`
- `attached_volumes` is commented out. The volume currently attached to `stats.usegalaxy.org.au` will be manually attached when we migrate.

### DIFF
--- a/stats-2025_playbook.yml
+++ b/stats-2025_playbook.yml
@@ -1,8 +1,10 @@
 - hosts: stats-2025
   become: true
   vars_files:
+    - group_vars/all.yml
     - group_vars/VAULT
     - secret_group_vars/stats_server_vault
+    - host_vars/stats-2025.yml
   pre_tasks:
     - name: install xfsprogs
       package:


hosts
```diff
@@ -237,6 +237,7 @@ qld-nvme ansible_ssh_host=203.101.231.166
 
 [stats_servers]
 stats.usegalaxy.org.au ansible_ssh_host=115.146.84.128
+stats-2025.usegalaxy.org.au ansible_ssh_host=115.146.87.139
```
cat-bro: Can this be called stats-2025 instead? the old stats has an awkward hostname. I see that you have tried not to double up on playbooks but maybe we need to until the existing VM is no longer needed.

host_vars/stats-2025.usegalaxy.org.au.yml
```diff
@@ -0,0 +1,356 @@
...
+  - 'mdu-research'
+
+retention_policy_8w_dbs:
+  - 'galaxy_etca'
+
+
+###### Grafana
+
+# set version (currently installed version on Ubuntu 20.04.5 LTS) so doesn't update Grafana when role is executed again
+grafana_version: "9.1.2"
```
cat-bro: Is this what you want on stats-2025? https://github.com/cloudalchemy/ansible-grafana/blob/b77b25607003c4cd9fd03091d03119d1dc8c2700/tasks/install.yml#L68-L76

hosts
jlqfab: Yep. will do!

cat-bro: The stats playbook also has patterns that we don’t normally use (which are valid patterns) of implicit variables files. Listing every file is a style we agreed on a while ago but it wasn’t done for this one - it means a person does not need to have memorised the structure of the hosts file to know where to find the variables.

jlqfab: Yep. I'll update the new playbook.

host_vars/stats-2025.usegalaxy.org.au.yml
jlqfab: I've updated (to the latest security release) and commented out `grafana_version` until this goes into production as we're on a blank VM right now.

host_vars/stats-2025.yml
```diff
@@ -0,0 +1,356 @@
...
+  - 'mdu-phl'
+  - 'mdu-research'
+
+retention_policy_8w_dbs:
+  - 'galaxy_etca'
+
+
+###### Grafana
+
+# TODO set version (installed version on Ubuntu 24.04 LTS) so doesn't update Grafana when role is executed again
```
cat-bro: sudo apt upgrade might do that anyway. I think you might have done something somewhere to stop it from upgrading

cat-bro (approved):

PR #2896, opened 2025-11-25
create one-off playbook for broken apt-key in dj-wasabi.telegraf

jlqfab:
This is the workaround for the `dj-wasabi.telegraf` role breaking on new VMs.

Rather than adding this code to every playbook that installs `dj-wasabi.telegraf`, it makes sense to have it as a standalone in one-offs. Then commenting out the role in the specific VMs playbook.

Alternatively, remove the actual install from here and ensure that `telegraf_repository_params` is set in each playbook that installs telegraf.

Either way works, Installing here just keeps it tightly tied to the workaround, while setting in each playbook ensures order of operations is unchanged.

@cat-bro I'm leaning towards the latter option, but documenting here for posterity.

### DIFF
--- a/one-offs/telegraf_install_with_fix_for_broken_apt-key.yml
+++ b/one-offs/telegraf_install_with_fix_for_broken_apt-key.yml
@@ -1,0 +1,47 @@
+- hosts: all
+  become true
+
+  pre_tasks:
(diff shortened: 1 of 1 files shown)


cat-bro:
This looks good.

An alternative would be setting telegraf_repository_params in all.yml and putting these tasks in the common role.

cat-bro (approved):

PR #2900, opened 2025-11-25
Add galaxy extentions role to staging/prod

neoformit:
Just tested again on dev

- Replace `webhooks` role with `galaxy_extensions` in staging/prod playbooks
- Move `toolmsg` vars to `group_vars/tool_messages.yml`

### DIFF
--- a/galaxy_playbook.yml
+++ b/galaxy_playbook.yml
@@ -6,6 +6,7 @@
     - group_vars/VAULT
     - group_vars/galaxyservers.yml
     - group_vars/galaxy_etca_slurm.yml
+    - group_vars/tool_messages.yml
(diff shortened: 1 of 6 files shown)


neoformit:
Ok, so apparently you can just do this:

```yaml
  connection: local
  become: false
```

That is supposed to revoke a `become: true` set at the play level. I've updated the role and tested with password-restricted sudo on my machine - it seems to work.

Are you happy to reinstall and give it another try?

cat-bro (approved):
Thank you @neoformit!

neoformit:
Wow this is the gift that keeps giving! I'll take a look, it must be possible...

neoformit:
Ok I have updated and tested locally with password-protected sudo.
Please merge #2959, then re-install and try again?

PR #2904, opened 2025-11-26
disable wal on galaxy-db

cat-bro:


cat-bro:
this was only added for the replicant db

PR #2926, opened 2025-12-05
nginx template changes for AAI

cat-bro:
This does not work yet so I’ll mark it as a draft. Pinging @uwwint into conversation and will request a review from Nuwan.

Error is `could not build map_hash, you should increase map_hash_bucket_size: 64`

### DIFF
--- a/templates/nginx/galaxy.j2
+++ b/templates/nginx/galaxy.j2
@@ -75,8 +75,8 @@ server {
 
     {% if aai_enabled %}
 
-    location = /authnz/keycloak/login {
-        proxy_pass http://galaxy_upstream;
+    location = /authnz/oidc/login {
+        proxy_pass http://galaxy;
         # Only add CORS for cross-origin callers we allow
         if ($cors_allow_origin != "") {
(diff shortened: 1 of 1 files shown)


uwwint:
@nuwang we want to login the user straight into galaxy from the outside. The CORS allows us to fetch a state from he OIDC provider so we can initiate a login from the outside.

cat-bro:
This is working now (I deployed it on dev and Uwe tested it). If it’s low priority and we have shown that it works on dev, can we park this until January?

jlqfab (commented):
This looks correct overall, with the CORS handling appropriately scoped and gated, and the use of a map for credentialed OIDC flows. I have no objections from an nginx or security perspective.

However, a couple of things I'm not sure of:
- Is `Access-Control-Allow-Credentials: true` required for this endpoint (i.e. are cookies or an `Authorization` header expected to be sent)? If not, it may not be necessary.
- What are the exact `Origin` values sent by browsers during the Auth0/OIDC login flow (e.g. expected hostnames, subdomains, ports)? These need to match the strict regex.

PR #2927, opened 2025-12-05
change to usegalaxy_eu.influxdbserver role for stats-2025 and update host vars

jlqfab:
needs to be tested.

### DIFF
--- a/stats-2025_playbook.yml
+++ b/stats-2025_playbook.yml
@@ -33,7 +33,7 @@
     - common
     - galaxyproject.nginx
     - geerlingguy.docker
-    - usegalaxy_eu.influxdbserver
+    - usegalaxy-au.influxdbserver
     - dj-wasabi.telegraf
     - grafana
 


cat-bro (dismissed):
looks good!

jlqfab:
@cat-bro I think I need a separate `requirements.yml` for stats-2025? I just realised that this will break our standard ansible when we do `make update-roles` as `usegalaxy_eu.influxdbserver` requires `ansible-core 2.17`.

jlqfab:
`requirements-2.17.yml` has roles and collections for running `stats-2025_playbook.yml` under ansible-core 2.17. The versions of roles and collections needs to be tested. It combines some of the Galaxy EU versions with some of our existing versions. This is still work in progress,

jlqfab:
This uses the galaxy AU fork of the `influxdbserver` role, and includes a custom policy variable, overridden in hosts vars to define the `two_months` policy for `galaxy_etca`.

Playbook is not yet tested with the latest changes.

PR #2938, opened 2025-12-12
increase worker count further in galaxy.usegalaxy.org.au.yml

cat-bro:
gunicorn has no `auto` option for this. The galaxy VM has 16 CPUs and runs TUS, gunicorn and gx-it-proxy. 

This is not a substitute for improving disk access times, but if more gunicorn workers can be stably run this could reduce the rate of web access errors.


jlqfab (approved):
Good idea! Thanks @cat-bro!

PR #2944, opened 2025-12-13
update galaxy handlers nginx for pulsar posts

cat-bro:
Add `proxy_read_timeout` for api location, make location specific to api/*/jobs. Without the proxy_read_timeout set high, galaxy has been hanging up on pulsar posts after 10 minutes (504 error) and it needs to stay on the hook for longer when the file is large. This seems to only be necessary when the upload store folder is not on the user data volume.

### DIFF
--- a/templates/nginx/galaxy-handlers.j2
+++ b/templates/nginx/galaxy-handlers.j2
@@ -51,13 +51,14 @@ server {
 
 {% endif %}
 
-    location /api {
+    location ~ ^/api/jobs/[^/]+/files$ {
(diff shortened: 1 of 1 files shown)


jlqfab (commented):
@cat-bro This looks good, but are there any other `/api` endpoints that need to be passed, as you're now blocking everything that isn't an `api/jobs/...`?

cat-bro:
Hi @jlqfab, this is the only api route needed on galaxy-handlers.

PR #2945, opened 2025-12-15
Labs engine expose Nginx internal files

neoformit:
This is required for the "Bootstrap Lab" feature to download files properly

### DIFF
--- a/roles/galaxy_labs_engine/tasks/main.yml
+++ b/roles/galaxy_labs_engine/tasks/main.yml
@@ -91,7 +91,7 @@
     - update
 
 - name: Set temp dir permissions
-  file: dest={{ labs_temp_dir }} mode=u=rwX,g=rwX,o=rwX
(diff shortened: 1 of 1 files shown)


roles/galaxy_labs_engine/tasks/main.yml
```diff
@@ -90,6 +90,12 @@
     - permissions
     - update
 
+- name: Set temp dir permissions
```
jlqfab: this should use permissions '02770' (set group ID with group rwx), owner and group should be galaxy.
Also, `labs_temp_dir` should not be in `/tmp', it should go into one of the galaxy temp directories (maybe `/mnt/tmp` which has other galaxy temp directories).

jlqfab: @neoformit `group=101` would be better if possible, especially given you've confirmed it works.

PR #2947, opened 2025-12-18
role for monitoring nfs disconnects

cat-bro:
Review the logs every 2 minutes to look for specific patterns related to NFS ‘not responding’ or ‘OK’ messages. This is running on galaxy-handlers and posting to slack when there is a message. It is posting to #dev-test-alerts (public channel) until we see how spammy it is. If it proves useful the channel can be updated to #alerts and the role can be added to the galaxy VM as well. I don’t think it’s worth putting this on the workers because there would be a lot of repetition (same message from 12 VMs in the event of an outage).


jlqfab (approved):
Thanks @cat-bro

PR #2958, opened 2026-01-05
Update production TPV version to 3.1.3

cat-bro:
Production galaxy is still using tpv version 2 and will need to upgrade to 3 along with release_25.1. This PR updates the linting workflow and updates the tpv_version in galaxyservers.yml to 3.1.3.

I experimented with verbose linting and an output artifact but I’ve dropped this because the pydantic warnings are not helpful in the absence of the temporary file of amalgamated rules as functions. Reducing the number of pydantic warnings is the subject of a different draft PR.

### DIFF
--- a/host_vars/dev.gvl.org.au.yml
+++ b/host_vars/dev.gvl.org.au.yml
@@ -5,7 +5,7 @@ add_hosts_handlers: yes
 common_logrotate_manage_rsyslog: true
 
 # total perspective vortex
-# Use these variables if running 'install-tpv' role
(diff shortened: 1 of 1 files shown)


host_vars/staging.gvl.org.au.yml
```diff
@@ -36,6 +36,8 @@ nfs_exports:
 galaxy_repo: https://github.com/galaxyproject/galaxy.git
 galaxy_commit_id: release_24.2
 
+tpv_version: "2.5.0"
```
nuwang:
```suggestion
tpv_version: "3.1.3"
```

PR #2962, opened 2026-01-09
Influx key rotation

cat-bro:
This makes a dearmored key from the new "influxdata-archive_compat-exp2029.key” and ensures that influxdata.list and telegraf.list use the same key

### DIFF
--- a/roles/common/tasks/influxdata_key.yml
+++ b/roles/common/tasks/influxdata_key.yml
@@ -28,11 +28,13 @@
     path: /etc/apt/sources.list.d/influxdata.list
   register: influxdata_list
 
-- name: Remove InfluxData Debian repo on Ubuntu
+- name: Remove InfluxData repo legacy lines
   ansible.builtin.lineinfile:
     path: /etc/apt/sources.list.d/influxdata.list
-    regexp: '^deb .*repos\.influxdata\.com/debian'
+    regexp: "{{ item }}"
     state: absent
+  with_items:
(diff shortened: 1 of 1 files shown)


roles/common/tasks/influxdata_key.yml
```diff
@@ -0,0 +1,71 @@
...
+    path: "{{ influxdata_key_path }}"
+    owner: root
+    group: root
+    mode: "0644"
+
+- name: Add InfluxData repository
+  ansible.builtin.apt_repository:
+    repo: >
+      deb [signed-by={{ influxdata_key_path }}]
+      https://repos.influxdata.com/ubuntu stable main
```
jlqfab: this needs to be `https://repos.influxdata.com/debian stable main` (ie `debian` rather than `ubuntu`), as per documentation at https://www.influxdata.com/blog/package-signing-key-rotation/ and https://repos.influxdata.com/ubuntu/ (manually tested yesterday).

cat-bro: Thanks. I’ll update this (and change the corresponding lineinfile removal earlier on)

cat-bro: This will still look a little different than the documentation because the keys are signed by "/usr/share/keyrings/influxdata-archive_compat.gpg” instead of "/usr/share/keyrings/influxdata-archive.gpg”

jlqfab (approved):
Thanks @cat-bro!

PR #3014, opened 2026-02-05
bwa_mem2: override tbvdb file size rules

cat-bro:
Not enough memory in xlarge input rule for bwa_mem2

Override each file size category in shared db. Hopefully this means that another rule “tpvdb_bwa_mem2_history_reference_rule” set there will still apply.


jlqfab (approved):
LGTM

PR #3065, opened 2026-03-16
Update README with migration script instructions

nuwang:
Added notes to include script modifications for legacy user prefixing and session invalidation.


cat-bro:
I think that the prefixing of purged usernames and the invalidating of user sessions could be separate scripts rather than being tacked on to anything that’s already being done. We just need to make sure running them at appropriate times is on the running sheet for the day.

PR #3067, opened 2026-03-16
override tpvdb fail rule for trinity

cat-bro:
There is a limit of 1GB in tpv-shared-db for trinity with advice for users to run their jobs with rnaspades instead. We tried it out but users are writing in, so this PR is to override the shared-db rule.


nuwang (approved):
I think this is much better. IMO, we should update shared db too.

PR #3074, opened 2026-03-19
Rename stop/start pulsar playbooks

cat-bro:
Fix inconsistent ‘yaml’ extension and give these playbooks `_playbook.yml` which allows us to make use of “make run".


jlqfab (approved):
Thanks @cat-bro

PR #3082, opened 2026-03-24
Update README.md in migrate_public_names

cat-bro:
Hi Nuwan, I’ve made an update to step (6).

My understanding of it is that in order to verify it we need to run it one more time in commit mode and observe that all rows of `migration_mapping_tmp` are unchanged after the repeated run.


nuwang (commented):
Since we've already run the migration script, I think even the first run will likely return 0 (since we enforce usernames for new signup via UI now), so a second run will do nothing yet again. Instead, I would just manually inspect the few, if any, rows that might change after running the migration script, and expect it to be close to, or exactly, zero.

PR #3098, opened 2026-04-01
tpv: make every config file a template

cat-bro:
We should be able to add variables to any TPV config file that we are managing (I’m not including the pulsar_scores file in in this because it is generated from db data).


nuwang (approved):
lgtm!

PR #3103, opened 2026-04-15
Update tools.yml

igormakunin2:
Increase memory allocation for dada2_assignTaxonomyAddspecies when input is >30 MB


cat-bro (approved):
Thank you @igormakunin2 !

PR #3108, opened 2026-04-16
Fix nginx postrotate task on production

cat-bro:
There has been a bug in this forever that has meant that the long_term_dir has been filling up with uncompressed log files alongside their zipped versions

This fix probably has implications for @neoformit ’s post_labs_log.sh cron job.


neoformit:
That's ok, `post_labs_log.sh` will be superseded by the log shipper anyway

PR #3112, opened 2026-04-22
dev gets chatbot

cat-bro:
As far as I can tell release_26.0 introduces ChatGXY in the sidebar in addition to the job error help (available via ladybug button for failed job). So far the job error help has not been on GA production.

This seems to be a package deal: setting `ai_model` in galaxy.yml enables both. I’ll make enquiries in matrix chats because I can see a use case for enabling only the job error help without ChatGXY.


cat-bro:
apparently you can choose different models for different purposes via inference services https://github.com/galaxyproject/galaxy/blob/release_26.0/lib/galaxy/config/sample/galaxy.yml.sample#L2984-L2999

PR #3114, opened 2026-04-23
Labs engine redis

neoformit:
This add a little infrastructure to the Labs Engine to make AI-generated labs work.

Because the OpenAI request takes a while, it has to run as a background job in a queue, which I've implemented with Redis.


cat-bro (approved):
LGTM

PR #3131, opened 2026-05-01
interactive tools need fewer volumes

cat-bro:
An IT running on slurm is using embedded pulsar and does not need the other volumes that slurm docker jobs would have.


cat-bro:
apologies if this nerfs ITs .. we’ll find out if it does but I don’t think they should need reference data or anything outside the working directory

PR #3132, opened 2026-05-01
disable algif playbook

cat-bro:
equivalent to the commands we ran on galaxy production. This might be handy for apollo too if we get the ssh key.


cat-bro:
chatgpt wrote the tasks for this one

PR #3137, opened 2026-05-04
Tiberius rules

TomHarrop:
Hi, I would like to test the Tiberius tool on GA a couple of times before releasing it into the wild.

The settings are copied from dorado...

I haven't done this in ages, LMK if I got anything wrong. Thanks!

### DIFF
--- a/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
+++ b/files/galaxy/dynamic_job_rules/production/total_perspective_vortex/tools.yml
@@ -4543,3 +4543,18 @@ tools:
     params:
       docker_enabled: true
       singularity_enabled: false
+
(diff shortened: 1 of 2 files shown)


cat-bro:
@TomHarrop if I’m following the wrapper, is this part `from_data_table=“tiberius_models”`only needed in some cases?

Thinking ahead, how big are tiberius models?

TomHarrop:
@cat-bro The intention of the data_table is to only allow users to select models that are actually in the container for the current tool version... the `conditional` filters the list of models based on a property of the input genome (Is it soft-masked). Does that make sense? 

The models are in the 100 MB range, but they are bundled in the container, hence no data manager.

cat-bro:
that is great!

cat-bro (approved):

PR #3151, opened 2026-05-11
Log shipping

neoformit:
Install vector log shipper on dev and prod to send Nginx logs to Nectar Object Store. For each server, this will send:

- Raw Nginx logs (shorter retention, maybe 1 month for debugging)
- Cleaned Nginx logs (longer retention, excludes noisy requests, spam and bots)
- Workflow invocations (infinite retention, small footprint)
- Tool runs (infinite retention, small footprint)

Logs are sent to S3 in batches when either 1 day or 100MB is reached.

:heavy_check_mark: Tested on dev

This PR also contains the workflows reporting refactored into a new `reporting-service` role, but that's currently untested and not used in any playbook. I can rebase that out if you want, but I figured it's doing no harm and will (hopefully) be deployed next sprint to its own VM. This includes the S3 retention policy, which is enforced by [s3_cleanup.py](https://github.com/neoformit/infrastructure/blob/log-shipping/roles/reporting-service/files/s3_cleanup.py) and defined in the `reporting-service` role defaults:

https://github.com/neoformit/infrastructure/blob/1a5de6e4f4062ffe50b3c036c37e4890bd4e031c/roles/reporting-service/defaults/main.yml#L40-L43

We should collectively review the filters for cleaning the Nginx logs:

https://github.com/neoformit/infrastructure/blob/1a5de6e4f4062ffe50b3c036c37e4890bd4e031c/roles/vector/templates/vector.yml.j2#L25-L69


cat-bro:
Can staging be skipped for this? I’m interested in removing the staging server at some point and less to maintain is better.

neoformit:
Sure, done

host_vars/stats.usegalaxy.org.au.yml
```diff
@@ -129,6 +129,7 @@ influxdb_databases:
   - 'dev'
   - 'mdu-phl'
   - 'mdu-research'
+  - 'workflow_invocations'
```
cat-bro: “workflow-invocations” is very specific. Is there a larger set of data we are interested in collecting that workflow-invocations could a measurement of?

roles/reporting-service/defaults/main.yml
```diff
@@ -0,0 +1,43 @@
...
+# Influx settings
+influx_url: "localhost"
+influx_db_utilisation: "utilisation-db"
+influx_db_queue: "queue-db"
+influx_db_stats: "stats-db"
+
+# Galaxy Database settings
+stats_db_server: localhost
+stats_database: galaxy
+stats_db_user: galaxy
```
cat-bro: reader (?)

roles/reporting-service/templates/workflows_reporting/env.j2
```diff
@@ -0,0 +1,11 @@
+GALAXY_ID_SECRET={{ host_galaxy_config.galaxy.id_secret }}
```
cat-bro: the there is a vault equivalent of this: vault_id_secret or something

cat-bro:
Thanks @neoformit . I cannot see where the reporting service role is run. Is there a playbook for that VM?

roles/reporting-service/files/s3_cleanup.py
```diff
@@ -0,0 +1,156 @@
...
+    """Delete a batch of objects from S3 (max 1000 per request)."""
+    for i in range(0, len(keys), 1000):
+        batch = [{'Key': k} for k in keys[i:i + 1000]]
+        client.delete_objects(
+            Bucket=S3_BUCKET,
+            Delete={'Objects': batch},
+        )
+
+
+def main():
```
nuwang, line 96: @neoformit It might be possible to simplify this if we just rely on bucket lifecycle policies. That is, we could either
1. configure object expiration policy on the swift bucket, or
2. it may also be possible through the vector s3 sink (by setting the appropriate headers - `x-amz-meta-*` headers may work with s3/swift).
3. Alternatively, there was also a dedicated swift sink which allows vector to set a ` X-Delete-After` header.

neoformit: Good thinking, I'll look into this. Thanks.

host_vars/stats.usegalaxy.org.au.yml
neoformit: God yes, sorry. This is old and I should have reviewed.

roles/reporting-service/defaults/main.yml
neoformit: FYI copied this from `slg.galaxy_stats` so you might wanna change that to reader too?

neoformit:
> Thanks @neoformit . I cannot see where the reporting service role is run. Is there a playbook for that VM?

@cat-bro I mention that in the description - I refactored this into a separate role but there's no host or playbook for it yet, I'll make another PR for that.

roles/reporting-service/defaults/main.yml
cat-bro: it's in the role's defaiults, not the specific variables we use for it.

roles/reporting-service/files/s3_cleanup.py
neoformit: 1. It looks like swift accepts `X-Delete-After:SECONDS`:
https://docs.openstack.org/swift/zed/api/object-expiration.html
But I can't see anything in the [aws_s3 sink docs](https://vector.dev/docs/reference/configuration/sinks/aws_s3/) that allows customizing request headers, so I think that's out.

1. I checked their list of sinks and swift is not there. Was it removed/deprecated?

1. So I guess that leaves us with bucket-level expiry which would not be a bad option, but it doesn't seem like Swift supports that.

So I don't think any of these are viable, sadly?

nuwang: Thanks for investigating. I used Google AI search and it confidently hallucinated a headers section on the s3 sink and a swift sink, neither of which is true. So a custom script seems like the sensible solution. Sorry about the noise.

neoformit: Haha no worries, it seemed like a great idea

neoformit:
I think this is just about ready, I've just seen in the dev logs that the filter is not quite working so please let me fix that before merge?

neoformit:
Ok good to go from my end

host_vars/stats.usegalaxy.org.au.yml
```diff
@@ -382,6 +384,15 @@ grafana_datasources:
     user: "{{ influxdb.grafana.username }}"
     secureJsonData:
       password: "{{ influxdb.grafana.password }}"
+  - name: "workflow_invocations"
+    type: influxdb
+    url: "{{ influxdb.url }}"
+    version: 2
+    editable: false
+    database: "workflow_invocations"
```
cat-bro: this section also reporting instead of workflow-invocations

staging_playbook.yml
```diff
@@ -50,6 +50,7 @@
     - postfix-mail-relay
     - role: usegalaxy_eu.walle
       tags: walle
+
```
cat-bro: please revert changes to this playbook

galaxy_playbook.yml
```diff
@@ -65,6 +65,8 @@
     - galaxyproject.tiaas2
     - role: usegalaxy_eu.walle
       tags: walle
+    - role: vector
```
cat-bro: please leave this out of galaxy playbook until it has been merged into dev

dev_playbook.yml
```diff
@@ -67,10 +67,13 @@
     - geerlingguy.docker
     - usegalaxy_eu.apptainer
     - dj-wasabi.telegraf
+    - vector
```
cat-bro: too many

cat-bro:
Hi @neoformit , I added a few comments. I’d rather put this in the dev playbook first before adding it to the production playbook.

Re: the vector role - what is the test and how does it work? there is a 38K file [roles/vector/tests/vector.log‎] that I can’t look at in github, I’d have to check out the branch to see what it is. There is also a smaller file containing access log lines that might be from our logs with real IPs. If this is needed, can those IPs be scrambled?

neoformit:
@cat-bro this has already been running on dev for a month or so. You want to deploy to dev yourself before putting on prod? Or is there another blocker?

> Re: the vector role - what is the test and how does it work?

Good question - I've just removed this, it was for manual testing of the vector config but I don't think it's necessary to keep.

cat-bro:
> his has already been running on dev for a month or so. You want to deploy to dev yourself before putting on prod

yes, thanks!

neoformit:
Ok, done!

neoformit:
@cat-bro is this ready?

cat-bro:
@neoformit could you please read through the files changed and address any remaining comments? Also, if there are extraneous files you are trying to commit (i.e. an .gitignore within a role) please remove those.

neoformit:
@cat-bro the only comment I see is from Nuwan - did you forget to complete a review?

I removed one .gitignore which was redundant but the other is keeping local testing stuff from being committed - it's not extraneous.

cat-bro:
nooooo!  Good catch @neoformit, I need to train my eyes for the ‘pending’ comments.

neoformit:
Thanks for the review @cat-bro good catches

cat-bro:
Thank you @neoformit , looking good. One last thing:

Adding a new mapping and sub-log in the nginx configuration:

```
map $request $log_workflow_invocation {
    default 0;
    ~*POST\ /api/workflows/.+/invocations 1;
}
```

```
    access_log {{ nginx_workflows_log }} combined if=$log_workflow_invocation;

```


Is this a antipattern given that the whole log is being shipped? Is there going to be a demand for more access log spin-offs?

neoformit:
Oh yes, these things are redundant now. Sorry this branch has been going for so long I've forgotten some of the older iterations of this project.

cat-bro (approved):
lgtm!

PR #3153, opened 2026-05-11
Add singularity volumes and run arguments to config

mthang:
added singularity_volumes and singularity_run_extra_arguments to dev tpv for ufcg_profile tool


mthang:
@cat-bro I just added two singularity arguments to dev tpv for ufcg profile to see if that's the issue

cat-bro:
can you change that just to the extra args? I do not believe the volumes need overriding, adding that there would not override them anyway

cat-bro (approved):

PR #3164, opened 2026-05-13
Add AGRF filesource

neoformit:
I also removed the decommissioned Griffith file source.

This will require Cat's webdav patch to work, but I tested and confirmed working locally.

### DIFF
--- a/templates/galaxy/config/file_sources_conf.yml.j2
+++ b/templates/galaxy/config/file_sources_conf.yml.j2
@@ -1,19 +1,3 @@
-{% if inventory_hostname.startswith('dev') %}
-- type: webdav
-  id: agrf_nextcloud
-  label: AGRF Nextcloud
-  doc: Import your files from AGRF's Nextcloud server. Configure access in User -> Preferences -> Manage Information
-  url: https://agrf-data.agrf.org.au
-  root: /remote.php/dav/files/${user.preferences['agrf_nextcloud_account|username']}/
-  login: ${user.preferences['agrf_nextcloud_account|username']}
-  password: ${user.preferences['agrf_nextcloud_account|password']}
-  writable: true
-  # Set the following settings to avoid loading entire files into memory
-  # useful when dealing with big files
-  use_temp_files: true
-  temp_path: /mnt/tmp
-{% endif %}
-
 - type: dropbox
   id: dropbox
   label: Dropbox
@@ -84,4 +68,18 @@
   secret: {{ vault_nectar_melbournegvl_cb_ec2_secret_key }}
   key: {{ vault_nectar_melbournegvl_cb_ec2_access_key }}
   writable: true
+
+- type: webdav
+  id: agrf_nextcloud
+  label: AGRF Nextcloud
+  doc: Import your files from AGRF's Nextcloud server. Configure access in User -> Preferences -> Manage Information
+  url: https://agrf-data.agrf.org.au
+  root: /remote.php/dav/files/${user.preferences['agrf_nextcloud_account|username']}/
+  login: ${user.preferences['agrf_nextcloud_account|username']}
+  password: ${user.preferences['agrf_nextcloud_account|password']}
+  writable: true
+  # Set the following settings to avoid loading entire files into memory
+  # useful when dealing with big files
+  use_temp_files: true
+  temp_path: /mnt/temp
 {% endif %}
--- a/templates/galaxy/config/file_sources_conf.yml.j2
+++ b/templates/galaxy/config/file_sources_conf.yml.j2
@@ -16,8 +16,8 @@
   id: agrf_nextcloud
   label: AGRF Nextcloud
   doc: Import your files from AGRF's Nextcloud server. Configure access in User -> Preferences -> Manage Information
-  url: "TBC"
-  root: "TBC"
+  url:  https://agrf-data.agrf.org.au
+  root: /remote.php/dav/files/
   login: ${user.preferences['agrf_nextcloud_account|username']}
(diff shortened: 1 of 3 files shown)


cat-bro:
If you change this to ‘AGRF filesource on dev’ and use the {%if } in the jinja templates to restrict it to dev, we can merge this straight away.  Best practice is to use ‘vault’ for user prefs secrets so that passwords are not stored in plain text in the galaxy database.

neoformit:
>Best practice is to use ‘vault’

If that's the case, I've added this to the other file sources with plain text passwords too.

And I consolidated the "if dev" conditionals into one block.

cat-bro:
> If that's the case, I've added this to the other file sources with plain text passwords too.

For anything that existed before the vault was created, we can’t switch to vault for those without breaking people’s existing creds, so please only use this for the file source you are adding.

neoformit:
Ahh ok sorry, undone that

templates/galaxy/config/file_sources_conf.yml.j2
```diff
@@ -1,11 +1,11 @@
...
-  label: Griffith ownCloud
-  doc: Import your files from ownCloud. Configure access in User -> Preferences -> Manage Information
-  url: "https://research-storage.griffith.edu.au"
-  root: "/owncloud/remote.php/nonshib-webdav"
-  login: ${user.preferences['griffith_owncloud_account|username']}
-  password: ${user.preferences['griffith_owncloud_account|password']}
+  id: agrf_nextcloud
+  label: AGRF Nextcloud
+  doc: Import your files from AGRF's Nextcloud server. Configure access in User -> Preferences -> Manage Information
+  url:  https://agrf-data.agrf.org.au
```
nuwang: Looks like some extra whitespace?
```suggestion
 url:  https://agrf-data.agrf.org.au
```

templates/galaxy/config/user_preferences_extra_conf.yml.j2
```diff
@@ -36,7 +38,7 @@ preferences:
             - name: token
               label: Personal Access Token used to create draft records and to upload files. You can manage your tokens at https://zenodo.org/account/settings/applications/
               type: secret
-              store: vault # Requires setting up vault_config_file in your galaxy.yml
+              store: vault
```
nuwang, line 41: IIRC, of you switch the store part-way though, the credentials will have to be re-entered. Flagging just in case users need a heads up.

nuwang (commented):
lgtm!

cat-bro:
I just noticed this part too. AGRF deals in some big data and I don’t think /mnt/tmp would be right for this

```
  # Set the following settings to avoid loading entire files into memory
  # useful when dealing with big files
  use_temp_files: true
  temp_path: /mnt/tmp
```

Is this relative to the web server (wherever web gunicorn is running)? I guess it must be.

@nuwang that would have been your original setting for Griffith owncloud. Do you think it’s accurate to how the webdav file source works?

nuwang:
Nice catch @cat-bro. That setting is apparently deprecated:
```
  # Deprecated. This option is ignored by the fsspec-based WebDAV file
  # source.
  #file_source_webdav_use_temp_files: true
```

templates/galaxy/config/file_sources_conf.yml.j2
```diff
@@ -1,16 +1,18 @@
...
-  label: Griffith ownCloud
-  doc: Import your files from ownCloud. Configure access in User -> Preferences -> Manage Information
-  url: "https://research-storage.griffith.edu.au"
-  root: "/owncloud/remote.php/nonshib-webdav"
-  login: ${user.preferences['griffith_owncloud_account|username']}
-  password: ${user.preferences['griffith_owncloud_account|password']}
+  id: agrf_nextcloud
+  label: AGRF Nextcloud
+  doc: Import your files from AGRF's Nextcloud server. Configure access in User -> Preferences -> Manage Information
+ url:  https://agrf-data.agrf.org.au
```
nuwang: @neoformit Sorry, my suggestion had deleted the whitespace from the wrong place. Have added the space in front of url and deleted the extra space after it now.
```suggestion
  url: https://agrf-data.agrf.org.au
```

cat-bro:
`temp_dir` will default to `new_file_path`

```
      file_source_temp_dir:
        type: str
        required: false
        desc: |
          Directory to store temporary files for file sources. This defaults to new_file_path if not set.
```
(https://github.com/galaxyproject/galaxy/blob/d3b3ab7288cd272bb1d709f3c63b0f9ea440bc06/lib/galaxy/config/schemas/config_schema.yml#L4423)

Could this be something like “{{ galaxy_tmp_dir }}/webdav_temp”? That would need to be added to `group_galaxy_extra_dirs` in galaxyservers.yml so that it would be created with galaxy perms. I wonder when this is used. In the case of a DATA FETCH job I would have expected it to go to the job working directory but there are other uses, export etc.

neoformit:
> Nice catch @cat-bro. That setting is apparently deprecated:
> 
> ```
>   # Deprecated. This option is ignored by the fsspec-based WebDAV file
>   # source.
>   #file_source_webdav_use_temp_files: true
> ```

I think we're just deploying this to a pyfilesystem patch for now, right? Otherwise fsspec will come in 26.1

neoformit:
Ok @cat-bro how's that?

cat-bro (approved):
Thanks @neoformit !

neoformit:
@cat-bro can this go on prod now? I'm not sure if we have technical blockers with the Galaxy fork?

PR #3177, opened 2026-05-21
Run dorado demux on pulsar-nci-test

neoformit:
It works! Though it produced an empty dataset, I'm not sure if that's expected behaviour. Will link the tool for the requesting user to test and verify once this is merged.


cat-bro (approved):

cat-bro:
Hi @neoformit , I’ve just updated the pulsar version on pulsar-nci-test, so it’s worth retrying the job and seeing if the output is any different.

neoformit:
Ok thanks

PR #3193, opened 2026-05-26
clean up alphafold rules

cat-bro:
the most common case (latest version of tool) was not the default case + there were a couple of rules that seemed redundant


cat-bro:
Hi @nuwang, I’m hesitating over this because my PR changes something that was matching versions via the tool key regex with something using helper functions on the version. Do you recommend one over the other?

nuwang:
I think the helper version is better, because it uses semantic version comparison. A lot less code too!

nuwang (approved):

PR #3211, opened 2026-06-11
Make 26.1 run on dev

cat-bro:
update galaxy role version

stop pinning gravity

galaxy role now installs dependencies with uv by default, I switched this off because of errors with the r-space package.

comment out ARGF file source because it’s not compatible with changes in galaxy (pydantic error) and galaxy would not start


cat-bro:
@neoformit heads up about the AGRF file source: it is temporarily commented out. Something in the schema with regard to this must have changed in the new release.

PR #3212, opened 2026-06-14
Deploy reporting service

neoformit:
This adds a new playbook for the reporting service, which has been run. Also fixes a number of issue identified with the reporting role and vector:

- New S3 object path format to arrange year/month/day as virtual folders, making it easier to navigate thousands of records
- Fixed vector's batching of S3 logs which were creating an object every second (had to  delete hundreds of thousands of tiny objects and start fresh)
- Fixed vector transforms which were producing zero records for workflow invocations
- Stop writing vector logs to stdout and redirect them to a local log file rotated to keep the last 50MB only for debugging purposes

I have run the vector updates on dev and it seems to be dispatching to S3 correctly. I disabled vector service on prod until the playbook runs again (hopefully with this!). If this hasn't been merged yet, can you please run prod playbook with `--skip-tags vector` to avoid restarting it with these issues unresolved.


reporting_playbook.yml
```diff
@@ -0,0 +1,19 @@
...
+- hosts: reporting_service
+  remote_user: ubuntu
+  become: true
+  vars_files:
+    - group_vars/all.yml
+    - group_vars/VAULT
+    - secret_group_vars/ssh_keys
+    - secret_group_vars/reporting_vault
+    - secret_group_vars/stats_server_vault
+    - host_vars/reporting.usegalaxy.org.au.yml
```
cat-bro, line 11: Could the host vars contain this block to send telegraf stats to the ancillary database?

https://github.com/usegalaxy-au/infrastructure/blob/master/host_vars/galaxy-aust-exports-2026.yml#L33-L41

neoformit: I guess so - this is to send the reporting VM's sys stats to Influx right?

cat-bro (approved):

cat-bro:
Looks great @neoformit !

PR #3215, opened 2026-06-17
upgrade galaxy to release 26.1

cat-bro:
Placeholder PR to add notes about upcoming release update

- What branch is will be used? Dev “release_26.1_au_dev” contains Marius M PR (to do: request backport) + a commit from Nuwan that is essential for login to work. Marius M has made a PR to python-social-auth to fix the same thing as Nuwan’s commit fixes. GA may or may not need to use a branch for release 26.1
[ Updated: release_26.1_au branch is the one ]

- One-off update required of authnz db records:
```
UPDATE oidc_user_authnz_tokens
SET provider = 'auth0'
WHERE provider = 'oidc';
```
- Updates to oidc_backends_conf will be needed to go with above change from ‘oidc’ to ‘auth0’ + AAI token expiry changes [ this is done ]

- Follow issue about increased db use: https://github.com/galaxyproject/galaxy/issues/22929. [ issue is resolved ]

PR #3217, opened 2026-06-17
aai data things conditional on galaxy version

cat-bro:
more secrets for qaai, put oidc_scope_prefix in galaxyservers.yml, condition oidc things and file sources on `__galaxy_major_version`

Production will need auth0_domain for release_26.1, ref #3215


templates/galaxy/config/file_sources_conf.yml.j2
```diff
@@ -89,3 +83,11 @@
 #   use_temp_files: true
 #   temp_path: "{{ galaxy_tmp_dir }}/webdav"
 {% endif %}
+
+{% if aai_enabled and __galaxy_major_version is version('26.1', '>=') %}
+- type: drs
```
cat-bro, line 88: @marius-mather is it OK to use the same entry for this for qaai, production galaxy instances

marius-mather: We'll need a different url_regex for each environment:

* dev: `^drs://aaidemo\.bioplatforms\.com/`
* qaai: `^drs://staging\.bioplatforms\.com/`
* prod: `^drs://bioplatforms\.com/`

PR #3247, opened 2026-06-29
Refactor reporting service scripts into separate repo

neoformit:
- Scripts were getting complex, so I moved them to https://github.com/usegalaxy-au/reporting-service to make them easier to manage/test
- In testing, I realized that the request path that Vector is collecting for a tool submission proxy is actually the tool form build request, not the actual job submission, so I have updated the regex to capture the correct path, `POST /api/tools `. This is more accurate and help us to cross-reference Nginx lines with DB records.

If this runs with the prod playbook tomorrow, I will delete all S3 logs to date and treat Thursday as day zero for S3 logs, and then include everything prior when I process prod's Nginx logs backlog with Vector.


neoformit:
Marking as draft while I test on reporting VM

cat-bro (approved):
LGTM!

cat-bro:
I ran vector role on production yesterday after merging this, so vector config has been updated since Tuesday middayish

neoformit:
Great thanks! I'll clean up the object store.

PR #3248, opened 2026-06-29
fix: wrong DRS URL for staging bioplatforms

marius-mather:
Sorry, got mixed up about which staging server we were using to test this, this should have the right value now.


amandazhuyilan:
I approve (although I have no rights here :) )

cat-bro (approved):

PR #3270, opened 2026-07-10
update qld mounts src and dest

cat-bro:
galaxy-aust-exports.genome.edu.au is the new VM that Justin set up for this purpose. The file paths are slightly more consistent than on the previous exports VM


cat-bro:
no more legacy VMs! (except for the staging cluster which is still on ubuntu 20.04 I think)

PR #3283, opened 2026-07-16
Update Vector backlog scripts

neoformit:
Nginx logs backlog is now all in Nectar Object Storage. This is just a few tweaks to the bash scripts that ran it - might be useful if we need to do something like this in future (like if we decide to send another log file to S3).

---

**Also** - removed `reporting` from `influxdb_retention_policy` list


cat-bro (approved):
LGTM!

PR #3344, opened 2026-08-17
extend client body timeout galaxy handlers

cat-bro:
Big files coming from pulsar are at risk of timing out if there is a break of more than a minute between successive reads. This has been happening often enough to risk jobs failings since there are only 5 retries per file.  Set `client_body_timeout` to 600s (default 60s).

https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_timeout

### DIFF
--- a/templates/nginx/galaxy-handlers.j2
+++ b/templates/nginx/galaxy-handlers.j2
@@ -51,6 +51,8 @@ server {
 {% if galaxy_config['galaxy']['nginx_upload_job_files_store'] is defined %}
 
     location /_job_files {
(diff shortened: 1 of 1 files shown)


nuwang (approved):
It might be worth trialling TUS uploads for this?

cat-bro:
The last I heard, .org switched from nginx-upload-module to TUS, but had a bad experience using TUS and moved back to nginx-upload-module. It does feel like there should be a resumable protocol here instead of “get it in one go or try again”. It’s possible that pulsar-relay changes what it is possible here (no idea).

PR #3353, opened 2026-08-20
add data42 as an object store backend for pulsar outputs only (revert of revert)

cat-bro:
Add data42 with weight 0 and job working directory on galaxy-handlers /pvol2 (2.5T vol, sharing space the with nginx upload store folder)

This change is because pulsars post a lot of things back to the job working directory on galaxy. Most of these are tiny but some are large files, 100GB+ in the case of hifiasm outputs.

I have tested this out on dev and it works, so I think it’s worth trying on production.

There is a TPV rule for _pulsar_destination that is commented out, uncommenting this will activate the setting of object store ID for pulsar jobs.


cat-bro:
I merged this by accident when I was trying to merge something else, so I reverted it and opened it again.

